PR #63, opened 2020-10-02
Initial configuration for bettercodehub

techsmyth:
set component level to 2; 
excluded generated code

### DIFF
--- a/.bettercodehub.yml
+++ b/.bettercodehub.yml
@@ -1,0 +1,6 @@
+component_depth: 2
+languages:
+- typescript
+- javascript
+exclude:
+  - /src/generated/.*


.bettercodehub.yml
```diff
@@ -0,0 +1,6 @@
+component_depth: 2
```
GhostOnTheFiber, line 1: Do we need to go further then 2, I see that `Client.Web/src/components/` has more then 2 depth?

GhostOnTheFiber (approved):
Looks good! Added one comment to clarify but if all good then go!

.bettercodehub.yml
techsmyth: 2 is enough, gets to the main components which is what this needs.

PR #65, opened 2020-10-05
Feedback to user on server connection status

atstojanov:

Adds support for error handling, when data is requested from the back end.
If an error occurs the request will be retried up to 25 times. Error 500 does not trigger retry.
If you navigate to not existing rout error 404 will be shown.
Different home page depending on if user is authenticated or not.
Errors are displayed in the upper right corner just under the login button.
Support global Error Boundary.

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added
- [x] I've read the contributing document https://github.com/cherrytwist/.github/blob/master/CONTRIBUTING.md
- [x] I've read and understand the Code of Conduct https://github.com/cherrytwist/.github/blob/master/CODE_OF_CONDUCT.md
- [x] I understand that any contributions or suggestions I made may make it into the actual code. I've read the License 


techsmyth:
image
Getting this error if no ".env" file set on the client. 

Related is that I get the following error when logging in: "Cannot read property 'replace' of undefined"

Reminds me that we also do need details of the .env file settings somewhere that all devs can easily access...

techsmyth (approved):
A lot in there!

Struggling to test properly. Moving to develop so the error handling can be used widely.

Issue #75, opened 2020-10-16
graphql endpoint is not configurable in case of "production" environment which is wrong


antst:
in src/index.tsx

```
const graphQLEndpoint =
  process.env.NODE_ENV === 'production'
    ? '/graphql'
    : process.env.REACT_APP_GRAPHQL_ENDPOINT || 'http://localhost:4000/graphql';
```

PR #76, opened 2020-10-16
Implementation of configuration variables processing at docker run time

### LINKED ISSUE #74: Handling of environment variables at deployment time
Currently, all environment variables which defines AAD/CT integration (REACT_APP_AUTH_CLIENT_ID etc) must be specified at docker build time. Although it is easy, it is not best case scenario. 
It would be much better if those variables would be set at deployment time (ie docker run -e REACT_APP_AUTH_CLIENT_ID=...).
Example of doing this is at https://www.freecodecamp.org/news/how-to-implement-runtime-environment-variables-with-create-react-app-docker-and-nginx-7f9d42a91d70/

antst:
Disclaimer: I am no way expert in node-js or ts, in fact  I nearly have no idea about both :)

This PR  fixes #74 and #75 for docker deployment.
It is not tested for local development and must be used rather as a guideline.

Tested to work within docker pipeline.


antst:
To work properly with local deployment, you will need to add call to env.sh script etc somewhere into the package.json.

src/utils/authConfig.ts
```diff
@@ -2,9 +2,9 @@
 // visit https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/configuration.md
 export const msalConfig = {
   auth: {
-    clientId: process.env.REACT_APP_AUTH_CLIENT_ID || '',
-    authority: `https://login.microsoftonline.com/${process.env.REACT_APP_AUTH_TENANT_ID}`,
-    redirectUri: process.env.REACT_APP_AUTH_REDIRECT_URI || 'http://localhost:3000',
+    clientId: window._env_.REACT_APP_AUTH_CLIENT_ID || '',
+    authority: `https://login.microsoftonline.com/${window._env_.REACT_APP_AUTH_TENANT_ID}`,
+    redirectUri: window._env_.REACT_APP_AUTH_REDIRECT_URI || 'http://localhost:3000',
```
atstojanov, line 7: Does not work in development mode.
image

PR #129, opened 2020-11-05
Multiple select with propositions. core component

OlegSubik:

- new input-multiple select. multi-selection support, + list of  propositions based on user input and possible results


techsmyth (approved):
Looks good!

src/components/core/MultipleSelect.tsx
```diff
@@ -0,0 +1,284 @@
...
+        cursor: 'pointer',
+      },
+    },
+    suggestionsTitle: {
+      marginTop: `${theme.shape.spacing(3)}px`,
+      marginBottom: `${theme.shape.spacing(2)}px`,
+      textTransform: 'uppercase',
+    },
+    suggestions: {
+      width: `${theme.shape.spacing(92)}px`,
```
Enderedne, line 110: Would avoid fixing the width of a component like that. Would suggest going for 100%, flex-grow or something similar.
Another idea is to incorporate media queries and have it responsive

src/components/core/MultipleSelect.tsx
```diff
@@ -0,0 +1,284 @@
...
+  };
+
+  const handleInputChange = e => {
+    const value = e.target.value.toLowerCase();
+
+    if (allowUnknownValues && e.key === 'Enter' && value !== '') {
+      const isAlreadySelected = selectedElements.find(el => el === value);
+      if (isAlreadySelected) return;
+
+      const newElements = elementsNoFilter.filter(el => el !== value);
```
Enderedne, line 217: Don't think elements should be changed when changing the input. The tags are there for quick access. I don't think we should be filtering them. Could confuse the user as to what we're filtering - tags or people?

src/components/core/MultipleSelect.tsx
```diff
@@ -0,0 +1,284 @@
...
+    label: {
+      width: '100%',
+      position: 'relative',
+      fontSize: '0.9vw',
+      whiteSpace: 'nowrap',
+      color: theme.palette.neutral,
+    },
+    selectContainer: {
+      position: 'relative',
+      width: `${theme.shape.spacing(92)}px`,
```
Enderedne, line 35: Same comment about responsiveness. Look below

Enderedne (changes requested):
Looks really nice! There are just a couple of small things I think need to be addressed before we can merge it.

src/components/core/MultipleSelect.tsx
OlegSubik: this line is for when a user hits enter with filled input, but yeah, I will change the filtering of the elements while typing in the input

Enderedne (approved):

PR #134, opened 2020-11-06
Navigation rings + back button

OlegSubik:

- Created navrings that displays the depth of the route

- Back button that will bring a user to a previous depth of router

### Additionan notes

- Navigation rings will be hidden lower than 'lg' breackpoint


techsmyth (approved):
Looks nice, and was working nicely for me. 

May be an issue with the admin screen + overlap but can pick that up later. Great to have this in!

PR #243, opened 2020-11-16
Styles update

OlegSubik:

- breadcrumbs hover color
- multiple select tags fonts
- opportunity page backgrounds
- projects texts overflow check


techsmyth (approved):
May want to move back to using the opportunity banner instead of the simple color - will pick that up later with Aldert...

PR #290, opened 2020-11-19
Warnings fix 

OlegSubik:
- Warnings fix + small ui adjustments

### DIFF
--- a/src/components/Admin/UserPage.tsx
+++ b/src/components/Admin/UserPage.tsx
@@ -99,7 +99,7 @@ export const UserPage: FC<UserPageProps> = ({ mode = EditMode.readOnly, user, ti
 
   const handleSave = (user: UserModel) => {
     // Convert UserModel to UserInput
-    const { id: userID, profile, memberof, ...rest } = user;
+    const { id: userID, profile, ...rest } = user;
     const userInput: UserInput = {
       ...rest,
       profileData: {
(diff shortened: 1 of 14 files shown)


src/components/Admin/UserPage.tsx
```diff
@@ -99,7 +99,7 @@ export const UserPage: FC<UserPageProps> = ({ mode = EditMode.readOnly, user, ti
 
   const handleSave = (user: UserModel) => {
     // Convert UserModel to UserInput
-    const { id: userID, profile, memberof, ...rest } = user;
+    const { id: userID, profile, ...rest } = user;
```
atstojanov, line 102: `memberof` should not be present in the user input, will fail the GraphQL validation.
Saving user will throw an error:
`Field "memberof" is not defined by type "UserInput".`

src/components/Admin/UserPage.tsx
```diff
@@ -130,6 +130,7 @@ export const UserPage: FC<UserPageProps> = ({ mode = EditMode.readOnly, user, ti
         },
       });
     } else if (isEditMode && user.id) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
       const { email, ...userToUpdate } = userInput;
```
atstojanov, line 134: The `email` extraction can be omitted. 
The back-end now checks if the `email` is different than the existing one, not only if it is in the input data. Please remove that and pass the `userInput` directly to the mutation.

atstojanov (dismissed):
There are comments that needs to be resolved.

src/components/Admin/UserPage.tsx
OlegSubik: done

src/components/Admin/UserPage.tsx
OlegSubik: done

techsmyth (approved):

PR #325, opened 2020-11-27
Community list/search results UI updates

OlegSubik:

- Upgraded cards in a list of community search results
- Added more brief info to the tiles
- Enhanced popups for user/group to present user/group info 
- Improved query for a search to retrieve parent of a group in order to display a level of creation and a parent name


techsmyth (approved):
Looks good! A big step forward.

A few minor updates: 
- ' Default Tags" ==> "Tags"
- If there are no tags then leave this field out.
- Location: Parent label to be the location as the label e.g. "Challenge: Balance The Grid". Ecoverse groups do not need to state a location.

techsmyth (approved):
Looks good!

PR #378, opened 2020-12-17
Aspects create

atstojanov:

- Possibility for admin to add new aspects
- Each new aspect might have the type that is not already created
- If every type is present, there will be no option to add a new one


techsmyth (approved):
Nice!

PR #381, opened 2020-12-18
Actors group create

OlegSubik:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: node_js
+node_js:
+- v12.18.3
+cache:
+  directories:
+  - node_modules
+install:
+- npm install
+script:
+- npm run-script lint:prod
+- CI=false npm run-script build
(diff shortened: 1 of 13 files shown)


techsmyth (changes requested):
The actor groups that can be created are defined by the settings from the server - which essentially gives the set of allowed actor groups. So behaviour should be:
- if all supported groups are present do not show the "add actor group option"
- if not all are present, then offer the user the option to create based on the list that is not yet present. The name is then hard coded

OlegSubik:
> The actor groups that can be created are defined by the settings from the server - which essentially gives the set of allowed actor groups. So behaviour should be:
> 
> * if all supported groups are present do not show the "add actor group option"
> * if not all are present, then offer the user the option to create based on the list that is not yet present. The name is then hard coded

done

techsmyth (approved):
Working well now I got the right branch!

PR #395, opened 2020-12-25
Challenge + Opportunity create/edit

OlegSubik:
- Edited Admin page, changed namings of actions
- Added possibility to create/edit opportunity/challenge
- Profile form (for opp or chall) is reused and using the same base and data output
- Form has a different view for Context edit and opp/chall management
- Profile page contains necessary validation + notifications of complete/error of an operation


techsmyth (approved):
Really nice set of updates!

There are a few minor edits that would like to see:
- TextID: this should be a smaller box (it is max 20 characters), and have a short description stating the following: "Unique textual identifier, used for URL paths. Note: cannot be modified after creation."
- Tagline should be before background i.e. the first of the editable fields.
- TextID should not be offered for editing but displayed as a read only string when updating a challenge / opportunity.

techsmyth (approved):

PR #449, opened 2021-01-21
Tagset/Reference components

atstojanov:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/components/Admin/Common/InputField.tsx
+++ b/src/components/Admin/Common/InputField.tsx
@@ -8,6 +8,7 @@ interface InputFieldProps {
   name: string;
   required?: boolean;
   readOnly?: boolean;
+  disabled?: boolean;
   type?: string;
   placeholder?: string;
   as?: React.ElementType;
@@ -19,6 +20,7 @@ export const InputField: FC<InputFieldProps> = ({
   name,
   required = false,
   readOnly = false,
+  disabled = false,
   type,
   placeholder,
   as,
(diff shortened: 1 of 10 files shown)


src/components/Admin/Common/InputField.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <Form.Control
+        name={name}
+        as={as ? as : 'input'}
+        type={type || 'text'}
+        placeholder={placeholder || title}
+        value={value}
+        onChange={field.onChange}
+        required={required}
+        readOnly={readOnly}
+        disabled={readOnly}
```
valentinyanakiev, line 43: disabled should be removed

src/models/Profile.ts
```diff
@@ -0,0 +1,5 @@
+export type Reference = {
```
valentinyanakiev, line 1: Move all profile entities to Profile.ts and use them from here (at the moment this type is not referenced anywhere and overlaps with the one in User.ts)

valentinyanakiev (changes requested):

src/components/Admin/Common/InputField.tsx
atstojanov: Not removed, but controlled separately.

valentinyanakiev:
Screenshot_2021-01-22_12-24-12
The offset between reference fields is bigger than the rest of the form on the admin user edit page.

valentinyanakiev (changes requested):
Code looks OK, just have a look at the offset between references.

valentinyanakiev (approved):

PR #484, opened 2021-03-02
Login Page

atstojanov:
- [ ] Missing: Register, Logout.
- [ ] Must check: Token Expiration.

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/components/LoginPage.tsx
```diff
@@ -0,0 +1,173 @@
...
+});
+
+const initialValues = {
+  username: '',
+  password: '',
+};
+const loginQuery = async (username: string, password: string) => {
+  return await axios
+    .post(
+      '/auth/login',
```
atstojanov, line 32: Must be read from configuration

src/models/Constantes.ts
```diff
@@ -0,0 +1,8 @@
+export const TOKEN_KEY = 'accessToken';
```
valentinyanakiev, line 1: Constants.ts not Constantes.ts - there is a typo

src/setupProxy.js
```diff
@@ -0,0 +1,18 @@
...
+  app.use(
+    '/graphql',
+    createProxyMiddleware({
+      target: 'http://localhost:4000',
+      changeOrigin: true,
+    })
+  );
+  app.use(
+    '/auth',
+    createProxyMiddleware({
```
valentinyanakiev, line 13: This configuration should be dynamic. If the endpoint of Simple Auth Provider is changed in the config - this is not reflected here, and I have a working example:
1. Docker compose the server project - it was changed to map 3003:3002 so Demo and Server project don't collide
2. Try to login with simple auth

It doesn't work and no error message is show, simply nothing happens when 'login' is clicked.

valentinyanakiev (changes requested):
Overall looks good, just fix the typo and create the proxy endpoints from the config. The rest looks OK to me.

valentinyanakiev (approved):

PR #485, opened 2021-03-06
improved feedback to user when client fails

techsmyth:


techsmyth:
@frankkienl think this should help make clearer what has happen

frankkienl (approved):
Great!

valentinyanakiev (approved):

PR #643, opened 2021-06-22
Adding lifecycle state display

pimotte:
Adding this to challenges and opportunities

Closes https://github.com/cherrytwist/client-web/issues/558

### Additional context

TODO:
- [x] Get styles in the right way. I've winged this a little bit. A pointer or two on this would be helpful. In particular there are 3 TODO's related to this that I left in the code.

### Governance 

- [x] Documentation is added. Doesn't change any development workflow, I'm not aware of a user guide or something that needs updating for this.
- [ ] Test cases are added or updated. I'd be happy to add one if you guys think it makes sense, but it might be reasonable to first look at the testing setup as a whole before tacking stuff on?

- I've read the contributing document https://github.com/cherrytwist/.github/blob/master/CONTRIBUTING.md.
- I've read and understand the Code of Conduct https://github.com/cherrytwist/.github/blob/master/CODE_OF_CONDUCT.md.
     https://github.com/cherrytwist/Coordination/blob/master/LICENSE.


src/components/ActivityPanel/StateActivityCardItem.tsx
```diff
@@ -0,0 +1,263 @@
+import React, { FC, useCallback, useState } from 'react';
+import { Maybe, Lifecycle } from '../../types/graphql-schema';
+import { createStyles } from '../../hooks/useTheme';
+import Typography from '../core/Typography';
+import { Button, Modal } from 'react-bootstrap';
```
atstojanov, line 5: Import Button from `../core/Button` so it is styled like the rest of the site.

atstojanov (changes requested):
Is there are way to change places of `in progress` and `abandoned`, because it looks like the `Abandoned` state is the straight path. Also the state change  `abandoned->in progress` is not visible`  image

Maybe something similar:
image
What kind of information will help you with that ordering?

pimotte:
> What kind of information will help you with that ordering?

I've thought about it for a bit, and I do think there's no way to get this without including more information. The longest path without loops is not a valuable heuristic as that could still swap complete and abandoned.

I guess specifiying a "happy path" with a sequence of nodes, or tagging it onto the machine defintion (does that allow arbitrary properties?) would work. 

Alternatively, we can cheat and hardcode this in the frontend, but that seems a little hacky.

Any preference?

techsmyth:
Re display of the states, think this needs to be split out into a separate item. 
Raised new card on the board: https://app.zenhub.com/workspaces/cherrytwist-5ecb98b262ebd9f4aec4194c/issues/cherrytwist/client-web/661

Can you address the other feedback point and then we can merge this into develop?!

pimotte:
I've fixed the styling issues and I'm totally good with postponing better layout. 

The only thing that I can't reproduce is the absence of the inProgress -> Abandoned transition, that's there for me in both firefox and chrome.

techsmyth (approved):

PR #921, opened 2021-08-04
Migrate button component to Material UI

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/themes/material/palette.ts
```diff
@@ -0,0 +1,22 @@
+import { PaletteOptions } from '@material-ui/core/styles/createPalette';
+import { defaultPalette } from '../defaults';
+
+// todo: how to insert the rest of the palette here?
+export const paletteOptions: PaletteOptions = {
```
Enderedne, line 5: You can do something like this 
```js
declare module '@material-ui/core/styles/createPalette' {
    export interface Palette {
        my_color: {
            main: string;
            light: string;
        };
    }
    export interface PaletteOptions {
        my_color: {
            main: string;
            light: string;
        };
    }
}
```

src/themes/material/typography.ts
```diff
@@ -0,0 +1,42 @@
+import { TypographyOptions } from '@material-ui/core/styles/createTypography';
+import { defaultTypography } from '../defaults';
+
+export const typographyOptions: TypographyOptions = {
```
Enderedne, line 4: If needed you can do the same thing for the typography as with the palette

src/themes/theme.ts
```diff
@@ -0,0 +1,64 @@
+export interface Palette {
```
Enderedne, line 1: This is the place where the type overrides need to happen - palettes, etc.

We should deprecate our theme-type-definitions and integrate/merge them with the material ones.

Once that is done most of the Theme imports and specified like below will become redundant.
From:
```jsx
import { Theme } from '../themes';

<Card
    classes={{
      background: (theme: Theme) => theme.palette.my_special_color,
    }}
/>
```
To:
```jsx
import { Theme } from '@material-ui/core';
<Card
    classes={{
      background: (theme) => theme.palette.my_special_color,
    }}
/>
```
We can close the review without changing the entirety of the Theme imports to material-ui, but the Theme definitions needs to be merged.

src/components/core/Button.tsx
```diff
@@ -222,25 +133,24 @@ const Button: FC<ButtonProps> = ({
...
+      className={clsx(className, inset && 'inset', small && 'small', block && 'block')}
+      classes={{
+        outlined: styles[variant],
+      }}
+      component={Link}
+      variant="outlined"
+      color={variant === 'primary' ? 'primary' : 'default'}
       {...props}
     >
       <Typography variant="button" color="inherit" weight="boldLight">
```
Enderedne, line 149: We need to drop the typography component inside the material button as well as the children. The material-ui buttons have their own props for passing icons - 
```jsx
<Button
  variant="contained"
  color="secondary"
  className={classes.button}
  startIcon={<DeleteIcon />}
>
  Delete
</Button>
```

Enderedne (changes requested):
Looks good, needs a little bit of work around the theme definition.

src/components/core/Button.tsx
hero101: I also think this way but changing that requires changes in 87 places of reference.
I think a compromise would be to support both of the approches and do the changes in a seperate story because it requires time and testing.

atstojanov (changes requested):
Cards bottom line disappear.
image

atstojanov (approved):

atstojanov (changes requested):
Translation not correct.
image

Enderedne (approved):

atstojanov (approved):

PR #971, opened 2021-08-16
Winston logging

valentinyanakiev:
Integrated winston logging with custom formatter - can be extended. Currently used a sample one from https://github.com/winstonjs/winston/issues/287. We can extend and adjust as needed. Moved all console.logs to use logger.info. Logger transports (currently only console) are disabled unless NODE_ENV='production'


hero101 (approved):

src/services/logging/winston/logger.ts
```diff
@@ -0,0 +1,63 @@
...
+// silent by default to be automatically deactivated
+// in production mode
+export const logger = winston.createLogger({
+  transports: [
+    new Console({
+      silent: false,
+      level: 'info',
+    }),
+  ],
+});
```
atstojanov, line 54: The comment doesn't reflect the actual logger creation. The comment says `silent by default` but the code says `silent: false`.

src/services/logging/winston/logger.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// don't log anything in production mode
+// probably should go further and return non
+// working logger function to reduce
+// execution time and improve speed results
+// on application
+if (process.env.NODE_ENV !== 'production') {
+  logger.transports.forEach(transport => (transport.silent = true));
+}
```
atstojanov, line 63: Comment says logging disabled for `production` but instead is disabled for every environment except `production`. 
Console logging must be enabled for `development`.

atstojanov (changes requested):

src/services/logging/winston/logger.ts
valentinyanakiev: Changed the logic to use the same env variable as in Alkemio Server - LOGGING_CONSOLE_ENABLED

src/services/logging/winston/logger.ts
valentinyanakiev: Changed the logic to use the same env variable as in Alkemio Server - LOGGING_CONSOLE_ENABLED

atstojanov (approved):

PR #1021, opened 2021-08-24
Tidy up env variables

atstojanov:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


techsmyth (commented):
This PR is good to go, fixes some legacy code.

However it is only a small part of the overall issue. What is the plan for the other items?

atstojanov:
> This PR is good to go, fixes some legacy code.
> 
> However it is only a small part of the overall issue. What is the plan for the other items?

They are still in use. The Sentry configuration can be moved to the server rest of the variables are just for build information and is normal to have them. The graphql endpoint can be preserved for development purposes, which is quite unlikely to be used.

Also the [aws-amplify deployment](https://create-react-app.dev/docs/deployment#aws-amplify) worth a try.

techsmyth:
OK, so we need to then extend the server to hold the sentry config, that would logically fit there. Will raise a separate issue for that. 

There are a couple of odd ones still in the .env.docker:
image

And regarding the .env file, is that only used for the .env.deployment script or does it also get picked up at runtime?

Finally, we should move the .env.deployment folder in under the .build folder, as it looks to be only part of the build - right?

Is the .env.production still needed?

The whole setup just feels too complicated, lots of environment variables + different usages.

techsmyth:
Just reflecting on this some more, I do think we need to keep the GRAPHQL_ENDPOINT env var. Sorry, missed its full usage this morning. The .env.production file can stay deleted. 

Otherwise follow up changes look good.

techsmyth (approved):
Looks good; nice to have this cleaner.

PR #1068, opened 2021-08-31
Fix: Scrolling

atstojanov:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/components/core/ScrollButton/ScrollButton.tsx
+++ b/src/components/core/ScrollButton/ScrollButton.tsx
@@ -1,0 +1,43 @@
+import { Box, Fade, useTheme } from '@material-ui/core';
+import { ReactComponent as ChevronUpIcon } from 'bootstrap-icons/icons/chevron-up.svg';
+import React, { useEffect, useState } from 'react';
+import Icon from '../Icon';
+import IconButton from '../IconButton';
+
+const ScrollButton = () => {
+  const [visible, setVisible] = useState(false);
+  const theme = useTheme();
+
+  const scrollToTop = () => {
+    window.scrollTo({
+      top: 0,
+      left: 0,
+      behavior: 'smooth',
+    });
+  };
+
+  useEffect(() => {
+    const toggleVisible = () => {
+      const scrolled = document.documentElement.scrollTop;
+      if (scrolled > 30) {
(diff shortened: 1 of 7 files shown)


src/components/core/ScrollButton/ScrollButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  useEffect(() => {
+    const toggleVisible = () => {
+      const scrolled = document.documentElement.scrollTop;
+      if (scrolled > 30) {
+        setVisible(true);
+      } else if (scrolled <= 30) {
+        setVisible(false);
+      }
```
hero101, line 26: why `30` exactly ??
also can be simplified to `setVisible(scrolled > 30)`

src/pages/About.tsx
```diff
@@ -52,6 +53,10 @@ const AboutPage = () => {
   const styles = useAboutStyles();
   const { data } = useServerMetadataQuery();
   const { t } = useTranslation();
+
+  const currentPaths = useMemo(() => [], []);
```
hero101, line 57: why is `useMemo` used here when it can be a constant?

src/root.tsx
```diff
@@ -68,6 +69,7 @@ const Root: FC = () => {
                     <NavigationProvider>
                       <UserProvider>
                         <App>
+                          <ScrollToTop />
```
hero101, line 72: Is this really the proper solution? Seems kind of strange.

hero101 (changes requested):

src/components/core/ScrollButton/ScrollButton.tsx
atstojanov: No reason at all.

src/pages/About.tsx
atstojanov: have to be memoised otherwise the breadcrumb will trigger infinity rerendering.

hero101: Even on constants?

atstojanov: It is possible to be constant if is defined outside the scope of the component.

hero101 (approved):

PR #1161, opened 2021-09-24
Fixed credential enum issues

valentinyanakiev:


hero101 (changes requested):
I believe the better approach is to keep the enums and translations 1:1, and modify to upper case if needed, e.g. using the URL params, which we decided to use lower case for.

PR #1166, opened 2021-09-24
Hub terminology; usability updates; bug fixes

valentinyanakiev:
Terminology
- Switched to using Hub instead of Ecoverse for a collection of Challenges (text / labels only for now, routes later)

Usability
- Provide users with a message when there are no users to select from when updating community / org / group membership
- Add tooltips to tags, titles and taglines on ecoverse / challenge / opportunity cards
- Added help icon with explanation to key fields on challenge / ecoverse / opportunity context

Bug fixes
- Allow global admins to access all routes
- Set of users for selecting on challenge / groups now comes from parent community
- Avoid continual refreshing of admin page on updating community membership


techsmyth (approved):
Nice set of stabilisations!

PR #1193, opened 2021-09-28
Admin: Opportunity Admins

atstoyanov:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/pages/Admin/Opportunity/OpportunityAdminAuthorizationPage.tsx
```diff
@@ -0,0 +1,65 @@
...
+      },
+    ],
+    [paths]
+  );
+
+  useUpdateNavigation({ currentPaths });
+
+  const { ecoverseId, opportunityId } = useOpportunity();
+
+  const { data, loading } = useOpportunityMembersQuery({ variables: { ecoverseId, opportunityId } });
```
hero101, line 31: you can use the generic `communityMembers` query rather than this one or stick to the same query for different entities, I am not sure which is the better approach.

src/components/Admin/managementData.ts
```diff
@@ -117,6 +117,15 @@ export const managementData = {
         { description: 'Updates', url: '/community/updates' },
       ],
     },
+    {
+      name: 'Authorization',
+      buttons: [
+        {
+          description: 'Opportunity admins',
+          url: '/authorization/admins',
```
hero101, line 125: if we are going to stop using the credential in the URL, isn't it better to just be `/admins` or `/authorization`?

hero101 (changes requested):

src/components/Admin/managementData.ts
atstoyanov: Keep it for now. Will be useful very soon.

hero101 (approved):

PR #1207, opened 2021-10-02
updated to work with application creation api for preserving order

techsmyth:
This PR works agains the server PR with the same number: https://github.com/alkem-io/server/pull/1429

This works against the updated server version, but does not actually send back up the sort order from the template. That is marked as a todo in the ApplyPage function.


techsmyth (commented):
Working nice with order preserved...

atstoyanov (approved):

valentinyanakiev:
Creating an application on dev:
Variable "$input" got invalid value { name: "What makes you want to join?", value: "123" } at "input.questions[0]"; Field "sortOrder" of required type "Float!" was not provided.

techsmyth:
Looks like the latest version of the client is not deployed?

PR #1234, opened 2021-10-07
Ecoverse page split

atstoyanov:
What's left:
- in `components` folder:
  - Admin - the whole admin part should be split into different composite  components/models/hooks etc.
  - Authentication - this is the KratosUI which has minimal dependency on the rest of the application (core/button and Delimiter + translation) - can be extracted in separate package.
- `OpportunityPage` should be split similar to the `Ecoverse` and `Challenge`
- `OrganizationPage` should be split similar to the `Ecoverse` and `Challenge`
  -  there are some composite components used by the organization page which can be moved into `composite/entities`


atstoyanov:
What's left:
- in `components` folder:
  - Admin - the whole admin part should be split into different composite  components/models/hooks etc.
  - Authentication - this is the KratosUI which has minimal dependency on the rest of the application (core/button and Delimiter + translation) - can be extracted in separate package.
- `OpportunityPage` should be split similar to the `Ecoverse` and `Challenge`
- `OrganizationPage` should be split similar to the `Ecoverse` and `Challenge`
  -  there are some composite components used by the organization page which can be moved into `composite/entities`

PR #1533, opened 2021-12-10
New user preferences

valentinyanakiev:


src/containers/user/UserNotificationsContainer.tsx
```diff
@@ -28,7 +28,9 @@ const limitNotifcationsTo = [
   UserPreferenceType.NotificationApplicationReceived,
   UserPreferenceType.NotificationApplicationSubmitted,
   UserPreferenceType.NotificationCommunicationUpdates,
+  UserPreferenceType.NotificationCommunicationDiscussionCreatedAdmin,
   UserPreferenceType.NotificationCommunicationDiscussionCreated,
+  UserPreferenceType.NotificationCommunicationDiscussionCreatedAdmin,
```
Comoque1, line 33: Looks like should be: `NotificationCommunicationUpdateSentAdmin`.

atstoyanov:
It might be better to reverse the logic and instead of a list of allowed values to populate a list of hidden values, so if a new value is added in the backend the value will be available on the frontend.

techsmyth:
image

Comoque1 (approved):

PR #1626, opened 2021-12-22
Image cropping - initial variant

atstoyanov:
This is an example of how images can be cropped on the client-side and sent to the backend using the existing API.
CropAvatar2

The image is not resized to 400x400 and the user should select an area that is less than 400x400 in order for the example to work.

What is left to be done so it can be used:

- [ ] Set the initial crop based on the image size in pixels - the current example uses "%" which is not taken into account when the final image is created. 
- [x] Resize the image to fit into 400x400 px.  [react-image-file-resizer](https://github.com/onurzorluer/react-image-file-resizer) can be used for the resizing or manual resize can be done during the construction of the final image (use the react-image-file-resizer source code)
- [ ] Create a component that can be reused for banners/backgrounds etc. The component should be able to:
  - [ ] accept aspect ration.
  - [ ] define maxHeight/maxWidth of produced image.
  - [ ] handy way of returning the final image. (react-image-file-resizer source code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@alkemio/client-web",
-  "version": "0.12.2",
+  "version": "0.12.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@alkemio/client-web",
-      "version": "0.12.2",
+      "version": "0.12.3",
       "license": "EUPL-1.2",
       "dependencies": {
         "@apollo/client": "^3.4.9",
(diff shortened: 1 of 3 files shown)


techsmyth:
really is a super win!

fyi, there is a new card up for the banners / backgrounds: https://github.com/alkem-io/client-web/issues/1628

so they can come out of the criteria for this PR and if you can get the % issue addressed lets prioritize getting this one into a new build

techsmyth (approved):

hero101 (approved):

PR #1657, opened 2022-01-07
Public and Private APIs+ invalid credentials handling

hero101:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/components/composite/layout/App/App.tsx
```diff
@@ -23,19 +22,21 @@ const App = ({ children }): React.ReactElement => {
...
+  const { services } = useServerMetadata();
+
+  useEffect(() => {
+    if (services.length) {
       console.table({
         clientName: process.env.REACT_APP_NAME,
         clientVersion: process.env.REACT_APP_VERSION,
-        serverName: data?.metadata.services[0].name,
-        serverVersion: data?.metadata.services[0].version,
+        serverName: services[0].name,
```
techsmyth, line 32: this is not safe; there could be multiple services and the order is not fixed

techsmyth (commented):
Great that this is working with the updated server api. 

When I am not logged in I am seeing a lot of errors show up in the graphql network explorer and also popups:
image

src/components/composite/layout/App/App.tsx
valentinyanakiev: We are going to have a much more granular service, already up to 3 back-end services, and this needs to be reviewed as an approach, whether that kind of table makes sense at all.

PR #1675, opened 2022-01-14
React router v6 upgrade

atstoyanov:
The React Router has been updated to V6.
Most routes have been updated to use the new syntaxis.
The `App.tsx` now works as a layout.

The `EcoverseTabsNew.tsx` and `EcoverseRouteNew.tsx` contain examples of how the tabs can be used as layouts. Unfortunately, I couldn't find a way to change the parent layout for nested routes which will lead to having 2 rows of tabs for challenges and 3 rows of tabs for opportunities.

One major bug. Navigating to ecoverse's challenges tab and challenge's opportunity tab is not working because of the existence of that route which is pointing to a completely different page.

```javascript
      <Route path={'/'}>
        <Route index element={<Navigate to={'dashboard'} />} />
        <Route path={'challenges/*'}>
          <Route
->          path={`:${nameOfUrl.challengeNameId}/*`}
            element={
              <ChallengeProvider>
                <CommunityProvider>
                  <ChallengeRoute paths={currentPaths} />
                </CommunityProvider>
              </ChallengeProvider>
            }
          ></Route>
        </Route>
->     <Route path={'*'} element={<EcoversePage paths={currentPaths} />} />
      </Route>
```
From the code is visible where the breaking is coming from. In this version of the tab navigation, we rely on the `TabPanel` components to switch between different views based on the location. 

The UserSettingsRoute is using a newer variant of the tab navigation which uses the generic `NavigationTabs` components and relies on the router to do the views switching. In my opinion, this approach is more clear and doesn't rely on some magic. The main problem is where to put the PageContainer that is delivering the data without the need to call it more than once, but this is for you to decide.

I can think of more solutions like:
- putting the `/challenges/*` route inside the EcoversePage
- renaming the `/challenges` to `/challenge`
but all of them are looking bad.


hero101 (approved):
Code reviewed and changes made by me
Tested by @Comoque1

PR #1739, opened 2022-02-08
Help to guide the user

valentinyanakiev:
I am not sure about some of the naming (e.g. helpText next to helperText) and the help on the Select control...


techsmyth (changes requested):
Tags are not being shown after being added:

image

On summary page the label is still default:
image

hero101 (approved):

techsmyth (approved):

PR #1978, opened 2022-04-14
Aspect card visual problems fixed

hero101:
### Describe the background of your pull request

What does your pull request do? Does it solve a bug (which one?), add a feature?

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added
- [ ] Test cases are added or updated

By submitting this pull request I confirm that:
- I've read the contributing document https://github.com/alkem-io/.github/blob/master/CONTRIBUTING.md.
- I've read and understand the Code of Conduct https://github.com/alkem-io/.github/blob/master/CODE_OF_CONDUCT.md.
- I understand that any contributions or suggestions I made may make it into the actual code. I've read the License 
     https://github.com/alkem-io/Coordination/blob/master/LICENSE.
- I understand that submitting the pull request requires agreeing to and signing the contributor license agreement.

PR #1980, opened 2022-04-14
Aspect delete moved to aspect settings tab

hero101:
Address this PR after #1978

### Governance 

- [ ] Documentation is added


src/pages/aspect/AspectSettingsPage.tsx
```diff
@@ -33,6 +34,9 @@ const AspectSettingsPage: FC<AspectSettingsPageProps> = ({ paths: _paths }) => {
       references: aspect?.references,
     };
 
+  const aspectIndex = resolved.pathname.indexOf('/aspects');
+  const contributeUrl = resolved.pathname.substring(0, aspectIndex);
```
me-andre, line 38: Did you try `navigate('..')`?

src/pages/aspect/AspectSettingsPage.tsx
```diff
@@ -58,25 +86,17 @@ const AspectSettingsPage: FC<AspectSettingsPageProps> = ({ paths: _paths }) => {
               <SectionSpacer />
               <EditVisualsView visuals={visuals} />
             </Box>
-            <Box sx={{ display: 'flex', justifyContent: 'end', marginTop: 2 }}>
+            <Box sx={{ display: 'flex', justifyContent: 'end', marginTop: 2, gap: t => t.spacing(1) }}>
```
me-andre, line 89: Let's name the variable `theme`?

src/pages/aspect/AspectSettingsPage.tsx
```diff
@@ -42,6 +46,30 @@ const AspectSettingsPage: FC<AspectSettingsPageProps> = ({ paths: _paths }) => {
...
       {(entities, state, actions) => {
         const visuals = (entities.aspect ? [entities.aspect.banner, entities.aspect.bannerNarrow] : []) as Visual[];
+        const btnDisabled = !aspect || !entities.aspect || state.updating || state.deleting;
+
+        const handleDelete = async () => {
+          if (!entities.aspect || !aspect) {
+            return;
+          }
+
+          await actions.handleDelete(entities.aspect.id);
```
me-andre, line 56: Should we `await`? If deletion takes a few seconds, a user may already start interacting - then he would be suddenly taken away from the page - which is suboptimal UX.

me-andre (changes requested):
Please address the comments

src/pages/aspect/AspectSettingsPage.tsx
hero101: `..` will navigate you just one step back - we want three steps instead.

src/pages/aspect/AspectSettingsPage.tsx
hero101: When the deletion takes a few seconds, and the user is navigated to a screen which will refetch the same data some errors are thrown because you tried to fetch the entity and it's sub-entities in the process and they may not exist.
I am opened for other suggestions.

src/pages/aspect/AspectSettingsPage.tsx
me-andre: `"../../.."`

me-andre: But if we just want to go to a specific route rather than "go up" (not sure if it's the case), I'd rather build a route from scratch.

src/pages/aspect/AspectSettingsPage.tsx
me-andre: Are you sure the errors are thrown? Because I assume most frameworks start cascaded deletion from the topmost entity.

me-andre: Otherwise any other user who opened the same page while some deletion is being performed would get the same errors 🤷‍♂️

src/pages/aspect/AspectSettingsPage.tsx
hero101: it happened to me several times, but now it stopped. I am removing it and we will see.

PR #1994, opened 2022-04-15
optimized users query on admin/users

hero101:
added loading animation to the data table

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/pages/Admin/User/UserListPage.tsx
+++ b/src/pages/Admin/User/UserListPage.tsx
@@ -1,41 +1,18 @@
 import React, { FC } from 'react';
 import { PageProps } from '../..';
 import ListPage from '../../../components/Admin/ListPage';
-import { SearchableListItem } from '../../../components/Admin/SearchableList';
-import { Loading } from '../../../components/core';
-import { useApolloErrorHandler } from '../../../hooks';
-import { useDeleteUserMutation, useUsersQuery } from '../../../hooks/generated/graphql';
+import UserListContainer from '../../../containers/user/user-list/UserListContainer';
 
 interface UserListPageProps extends PageProps {}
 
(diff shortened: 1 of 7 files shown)


src/pages/Admin/User/UserListPage.tsx
```diff
@@ -1,41 +1,18 @@
...
-  };
-
-  if (loading) {
-    return <Loading text={'Loading Users ...'} />;
-  }
-  return <ListPage data={userList} paths={paths} onDelete={handleDelete} />;
+  return (
+    <UserListContainer
+      component={({ userList, onDelete, loading }) => (
+        <ListPage data={userList} paths={paths} onDelete={onDelete} loading={loading} />
```
me-andre, line 12: Please use children function for that use case.
```
<UserListContainer>
  {({ userList, onDelete, loading }) => (
    <ListPage data={userList} paths={paths} onDelete={onDelete} loading={loading} />
  )}
</UserListContainer>

**Review by me-andre — CHANGES_REQUESTED** [review]

File: src/pages/Admin/User/UserListPage.tsx (hunk 1 above)
Comment (me-andre) [reply src/pages/Admin/User/UserListPage.tsx:12]: `component=` signature is only for the exact match between the provided props and the child component props.

Comment (hero101) [reply src/pages/Admin/User/UserListPage.tsx:12]: > `component=` signature is only for the exact match between the provided props and the child component props.

Can we forbid using a child function in the component prop?
Maybe just `ComponentClass` instead of `ComponentType`?

Comment (me-andre) [reply src/pages/Admin/User/UserListPage.tsx:12]: Well, a component is a function. I think we can take another direction and just pass a component as a child fn:
```
<Container>
  {Component}
</Container>

me-andre (approved):

PR #2012, opened 2022-04-22
Entity Settings pages have explanation headers

me-andre:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


src/i18n/en/translation.en.json
```diff
@@ -1089,5 +1089,13 @@
   "associates-view": {
     "less": "show less",
     "more": "...plus {{count}} more"
+  },
```
techsmyth, line 1092: I think that for the text to have meaningful content then it has to be specific to each tab that is shown. So I think there will need to be a translation section for each entity.tab. 

So there would need to be a hubs, users, organizations, authorization section underneath the admin level etc., with tabs below each of those. Make sense?

techsmyth (changes requested):
Looks like a clean update. Nice to see it getting with less code  :)

Main feedback is that I think for these labels to have meaning then we will need to have them specific to each tab.

The organization tabs were for me not working properly after this update. Profile tab name was Context, and other tabs gave a 404:
image

src/i18n/en/translation.en.json
me-andre: I updated the structure.
So now, whenever one comes up with a custom explanation for a tab under an entity, he just copies the part of the structure from "generic" under e.g. "hub" and edits it there:
```
    "hub": {
      "profile": { "explanation": "This is a very specific message about the Hub Profile." }
      ...
    }

```
Does it work for you, @techsmyth ?

me-andre:
The bug with navigation under Organization is addressed.

src/i18n/en/translation.en.json
techsmyth: Yes, thanks - this works...

techsmyth (approved):
Working nicely.

PR #2224, opened 2022-06-22
Upgrade Sentry to 7.2.0

ccanos:

Upgrade Sentry to 7.2.0 and deal with small breaking changes

### Additional context
https://github.com/getsentry/sentry-javascript/blob/master/CHANGELOG.md
* Removed critical Severity level. (#5032)
* Deprecated Severity enum for bundle size reasons. [Please use string literals instead](https://github.com/getsentry/sentry-javascript/blob/master/MIGRATION.md#severity-severitylevel-and-severitylevels). (#4926)

### Governance 

- [ ] Documentation is added


src/services/logging/sentry/log.ts
```diff
@@ -4,7 +4,7 @@ export const error = (
   error: Error,
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   setup: (scope: Sentry.Scope) => void = () => {},
-  severity: Sentry.Severity.Critical | Sentry.Severity.Fatal | Sentry.Severity.Error = Sentry.Severity.Error
+  severity: 'fatal' | 'error' = 'error'
```
valentinyanakiev, line 7: Looking at the link you have added, it will be cleaner to use, e.g. for the warning:

```
export const warningLevel: SeverityLevel = 'warning';
```

And then:

```
export const warn = (warning: string) => {
  Sentry.withScope(scope => {
    scope.setLevel(warningLevel);
    Sentry.captureEvent({ message: warning });
  });
};
```

Naming and placing might need to be changed, but the idea is to avoid using string literals directly and to wrap them up in constants (and that's actually what they suggested:

Severity, SeverityLevel, and SeverityLevels
We deprecated the Severity enum in @sentry/types and it will be removed in the next major release. We recommend using string literals (typed as SeverityLevel) to save on bundle size.

// New in 6.17.5:
import { SeverityLevel } from '@sentry/types';

const levelA = "error" as SeverityLevel;

const levelB: SeverityLevel = "error"

// Before:
import { Severity, SeverityLevel } from '@sentry/types';

const levelA = Severity.error;

const levelB: SeverityLevel = "error"

valentinyanakiev (changes requested):

src/services/logging/sentry/log.ts
valentinyanakiev: The benefit is that you are actually still using Sentry types. If they change something in the future (possible, albeit unlikely), e.g. drop info level, you will immediately see that upon SDK update when you build the solution. With direct string literals, untyped, you will end up seeing this error in production.

ccanos: Good, understood,
I'll push something in a few minutes

valentinyanakiev (approved):

Issue #2180, opened 2022-06-14
Max length notification comments


larssondenise:
## Description

As a User I want to see what the maximum amount of words is I can use when submitting a comment under Aspects (als Updates).
As a User I don't want to loose the text I have written after trying to submit it.

## Acceptance criteria

- [ ] User interface for showing the maximum amount of words that can be used
- [ ] keep the text in the box after error message

## Additional Context
Email from UWV .... 
Translation: 
I just entered a whole wall of text regarding my request for help. When I pressed post comment I got the message that apparently there were too many characters. Also, my text, which I spent quite a bit of time and attention in, has disappeared!

I think you would like to have this info to improve the platform. Maybe an idea to indicate in advance how many characters you can put in? I myself have given up, so will not try to reproduce the same text again. This really sucks.
![Schermafbeelding 2022-06-14 om 09.43.16.png](https://images.zenhubusercontent.com/6200e10a5561c893d9e84591/7a359e72-ec2f-4633-8ce4-2630adf673f1)

valentinyanakiev:
We have a limit on characters, not words (what the user actually requested). Needs discussion.

PR #2291, opened 2022-07-04
Add length counter and length limit to the MD editor #2180

ccanos:

Add length counter and limit to the MD editor #2180

### Governance 

- [ ] Documentation is added


src/components/composite/common/Discussion/PostComment.tsx
```diff
@@ -26,10 +28,13 @@ const PostComment: FC<PostCommentProps> = ({ onPostComment, title, placeholder }
     post: yup.string().required(t('forms.validations.required')),
   });
 
-  const handleSubmit = async (values: formValues, _helpers: FormikHelpers<formValues>) => {
+  const handleSubmit = (values: formValues, _helpers: FormikHelpers<formValues>) => {
     if (onPostComment) {
-      await onPostComment(values.post);
-      _helpers.resetForm();
+      return (onPostComment(values.post) as Promise<FetchResult<void>>)?.then(result => {
```
ccanos, line 33: this is a `void | Promise<FetchResult<void>>` casted to just `Promise<FetchResult<void>>`, haven't found a way to make it cleaner... just the `?` doesn't work and gives me the error "Type `void | Promise...` doesn't have the function `then`

src/context/Discussions/DiscussionProvider.tsx
```diff
@@ -143,7 +144,8 @@ const DiscussionProvider: FC<DiscussionProviderProps> = ({ children }) => {
           message: post,
         },
       },
-    });
+      onError: handleError,
+    })) as Promise<FetchResult<void>>;
```
ccanos, line 148: Here I'm casting a `Promise<FetchResult<PostDiscussionCommentMutation, Record<string, any>, Record<string, any>>>` to `Promise<FetchResult<void>>` because I don't want to have to drag those imports, `Record`, `PostDiscussionCommentMutation` all the way up the dependent classes..
I only need to access the `FetchResult` to see if it has errors or not. But I'm not sure if this is the cleanest way.

src/components/composite/common/Discussion/PostComment.tsx
```diff
@@ -26,10 +28,16 @@ const PostComment: FC<PostCommentProps> = ({ onPostComment, title, placeholder }
     post: yup.string().required(t('forms.validations.required')),
   });
 
-  const handleSubmit = async (values: formValues, _helpers: FormikHelpers<formValues>) => {
+  const handleSubmit = (values: formValues, _helpers: FormikHelpers<formValues>) => {
     if (onPostComment) {
-      await onPostComment(values.post);
-      _helpers.resetForm();
+      const result = onPostComment(values.post);
+      if (result) {
```
hero101, line 34: this will never be `undefined | null`. Also await it instead

hero101 (changes requested):

src/context/Discussions/DiscussionProvider.tsx
ccanos: I can't do that because I need the FetchResult outside to see if there were any errors. 
I just don't need the rest of the _Fetched Result_ properties.

ccanos: Very good one. Thanks

hero101 (approved):
My suggestions were addressed.

PR #2545, opened 2022-09-02
Client restructuring part #1

valentinyanakiev:
First step of client restructuring:

- moved multiple views, hooks, models, routers, pages etc from src/[views|hooks|routers...] to src/domain/[entity/routers/...
- ended up with a structure similar to:
image
- started moving core / common logic to src/core src/common
- moved 80-90% of graphql folders to src/domain/entity/graphql
- updated codegen.yml (will need further updates down the road when files are moved around)
- integrated & (light) tested current develop branch into client-2405

I have installed Webstorm locally to handle the refactoring better...
I suggest we move this forward and test+merge it so we don't get more divergance. It's utterly impossible to be reviewed anyway...


techsmyth (approved):
Awesome to have this refactoring in place. Merging now so we can get everyone onto this baseline.

PR #2581, opened 2022-09-08
Scrollbar color in Firefox

ccanos:

I've done a little investigation, and looks like customizing scrollbars the way we do it is not supported by firefox because it's not a standard yet.
https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

There is a supported way, but it's way more limited:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars
Only allows to change the color. The width can be changed too but only between a set of predefined values...

I've played a bit with it on this fiddle and got to have an 'okay' result...
Just needs this change
`scrollbar-color: #068293 transparent`

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

### DIFF
--- a/src/root.tsx
+++ b/src/root.tsx
@@ -19,6 +19,9 @@ import { ConfigProvider } from './config/context/ConfigProvider';
 
 const useGlobalStyles = makeStyles(theme => ({
   '@global': {
+    '*': {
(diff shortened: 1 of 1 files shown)


hero101 (approved):

hero101:
Your commit is not signed.

PR #2616, opened 2022-09-16
Wrappers

valentinyanakiev:
- PR prefixing core composite components created from MUI classes and clashing 1:1 with MUI names
- 1 PR per component
- Fixed eslint version due to a well-known issue in WS - https://youtrack.jetbrains.com/issue/WEB-57089
- Removed unused components

To review: 
- Go commit by commit, not file by file. 
- Double check whether all the components removed should be removed (they are not used, but were they intended to be used?)
- Check whether all the wrappers are really wrappers. I am using the notation of `Wrapper` as classes with the same purpose as their `MUI` counterparts, as well as the same name
- There are no functional changes in any files, just imports and errors / warnings auto-fixes
- Added documentation - by no means extensive, but if you open development-setup.md and target-development-setup.md and compare the folder structures, the intention + direction where we want to go gets quite clear.


hero101 (approved):

hero101:
closing because cla does not report

hero101 (approved):

PR #2649, opened 2022-09-22
Applied agreed upon eslint rules and applied fixes

valentinyanakiev:
- applied rules in the story
- ran `npm run eslint:fix`
- fixed issues with translation (it was invalid json)

no additional work required based on this rules


src/common/components/core/TextInput.spec.tsx
```diff
@@ -45,7 +45,7 @@ describe('TextIpnut component', () => {
 
   test('TextArea onChange event', async () => {
     // act
-    render(<TextArea onChange={onChange} value={value} label={label} inset={true} small={true} disabled={false} />);
```
alstojanovic, line 48: `disabled` has a default value of `false`, this is redundant.

src/common/components/core/TextInput.spec.tsx
```diff
@@ -55,7 +55,7 @@ describe('TextIpnut component', () => {
 
   test('TextInput disabled="false" input', async () => {
     // act
-    render(<TextInput onChange={onChange} value={value} label={label} inset={true} small={true} disabled={false} />);
+    render(<TextInput onChange={onChange} value={value} label={label} inset small disabled={false} />);
```
alstojanovic, line 58: `disabled` is false by default.

src/common/components/core/TextInput.spec.tsx
```diff
@@ -65,31 +65,31 @@ describe('TextIpnut component', () => {
 
   test('TextArea has value', async () => {
     // act
-    render(<TextArea onChange={onChange} value={value} label={label} inset={true} small={true} disabled={false} />);
+    render(<TextArea onChange={onChange} value={value} label={label} inset small disabled={false} />);
```
alstojanovic, line 68: disabled is false by default.

src/common/components/core/TextInput.spec.tsx
```diff
@@ -65,31 +65,31 @@ describe('TextIpnut component', () => {
...
 
     // assert
     expect(screen.getByRole('textbox')).toHaveValue(value);
   });
 
   test('TextInput has value', async () => {
     // act
-    render(<TextInput onChange={onChange} value={value} label={label} inset={true} small={true} disabled={false} />);
+    render(<TextInput onChange={onChange} value={value} label={label} inset small disabled={false} />);
```
alstojanovic, line 76: disabled is false by default.

src/common/components/core/TextInput.spec.tsx
```diff
@@ -65,31 +65,31 @@ describe('TextIpnut component', () => {
...
 
     // assert
     expect(screen.getByRole('textbox')).toHaveValue(value);
   });
 
   test('TextArea has label', async () => {
     // act
-    render(<TextArea onChange={onChange} value={value} label={label} inset={true} small={true} disabled={false} />);
+    render(<TextArea onChange={onChange} value={value} label={label} inset small disabled={false} />);
```
alstojanovic, line 84: disabled is false by default.

src/common/components/core/TextInput.spec.tsx
```diff
@@ -65,31 +65,31 @@ describe('TextIpnut component', () => {
...
 
     // assert
     expect(screen.getByText(label)).toBeInTheDocument();
   });
 
   test('TextInput has label', async () => {
     // act
-    render(<TextInput onChange={onChange} value={value} label={label} inset={true} small={true} disabled={false} />);
+    render(<TextInput onChange={onChange} value={value} label={label} inset small disabled={false} />);
```
alstojanovic, line 92: disabled is false by default.

alstojanovic (changes requested):

alstojanovic (changes requested):
`disabled` prop needs to be removed completely. Boolean props generally work on the principle they are `false` by default in which case they are not added as a prop. Otherwise, if we want them to be `true`, we add them as a prop.

src/common/components/core/TextInput.spec.tsx
valentinyanakiev: This will actually require further changes and feels out of scope. `npm run test:coverage' fails with the suggested change.

src/common/components/core/TextInput.spec.tsx
valentinyanakiev: This will actually require further changes and feels out of scope. `npm run test:coverage' fails with the suggested change.

src/common/components/core/TextInput.spec.tsx
valentinyanakiev: This will actually require further changes and feels out of scope. `npm run test:coverage' fails with the suggested change.

src/common/components/core/TextInput.spec.tsx
valentinyanakiev: This will actually require further changes and feels out of scope. `npm run test:coverage' fails with the suggested change.

src/common/components/core/TextInput.spec.tsx
valentinyanakiev: This will actually require further changes and feels out of scope. `npm run test:coverage' fails with the suggested change.

alstojanovic (approved):

PR #2662, opened 2022-09-26
Hot fix for login looping

techsmyth:


valentinyanakiev:
Hotfix branched off develop. Should be from master.

PR #2667, opened 2022-09-26
Search card refresh

hero101:
- The search cards version does not represent 1:1 the mockups because of dimension constraints.
I tried matching the styling and structure of the mockup.
- I have also provided two class diagrams so the architecture visibility is improved and further extension of the cards is easier.

**I suggest only critical or minor issues be addressed in this PR, and bigger changes be addressed later.**


src/pages/Search/SearchPage.tsx
```diff
@@ -40,7 +47,7 @@ const contributorFilterConfig: FilterConfig = {
 const entityFilterConfig: FilterConfig = {
```
ccanos, line 47: `journeyFilterConfig`

src/common/themes/default/defaults.ts
```diff
@@ -26,6 +26,17 @@ export const theme: ThemeOptions = {
     },
   },
   cards: {
+    search: {
+      width: 264,
```
ccanos, line 30: The constants I'm adding here are normally passed through `theme.spacing`, for example in simpleCard down there the width is 32 * spacing(8px) = 256px. 
I see that your height for example is not a multiple of 8 which is a problem to do what I'm saying, so I don't know what I'd do...

- If it's okay to make cards a few pixels bigger or smaller to match the 8px spacing grid, great.
- If graphically they are pixel perfect, I've calculated 394 = 49.25, so contributor.height would be 49.25, that's not a terribly ugly code...
- And if you don't agree, just leave it like this, it is fine.

src/domain/shared/components/search-cards/base/SearchBaseCard.tsx
```diff
@@ -0,0 +1,115 @@
...
+import CardContent from '@mui/material/CardContent';
+import CardActions from '@mui/material/CardActions';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import TagsComponent from '../../TagsComponent/TagsComponent';
+import { ClampedTypography } from '../../ClampedTypography';
+import { RouterLink } from '../../../../../common/components/core/RouterLink';
+import Link from '@mui/material/Link';
+
+const MATCH_TERMS_HEIGHT = 50;
```
ccanos, line 14: maybe these could go in the theme constants file too

src/pages/Search/SearchResultCardChooser.tsx
```diff
@@ -1,42 +1,278 @@
...
-  if (!card) {
+  if (card === undefined) {
     throw new Error(`Unrecognized result typename: ${result.__typename}`);
   }
 
   return card;
 };
 export default SearchResultCardChooser;
+
+const useHydrateCard = (
```
ccanos, line 74: I'd put this in a separate file.

ccanos (approved):
Just minor comments

src/common/themes/default/defaults.ts
hero101: Using the spacing requires the whole application to be based on this spacing constant.
The contents of the cards should be then dynamically calculated - lines truncated, wrapped, etc.
Having hardcoded dimensions is more reliable at this point imo.

src/domain/shared/components/search-cards/base/SearchBaseCard.tsx
hero101: I really don't know. These are constants are related to the base component itself. Which will not be used directly.

ccanos (approved):

PR #2882, opened 2022-11-04
Import Templates dialog

ccanos:
This is a functional demonstrator to demonstrate the flows around importing templates from an InnovationPack.

The functionality is only available to Global Admins, Global Hub Admins - so the fact the UI is rough is fine for that audience for now.

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -992,7 +992,7 @@
             "innovation-templates": "Innovation Flow Templates",
             "delete-confirmation": "Are you sure you want to delete {{template}} from the hub? You can also edit the information of this template.",
             "import-innovation-packs": {
-              "title": "Import an Innovation Pack"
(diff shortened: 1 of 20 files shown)


techsmyth:
Nice :) Import button is showing up but am seeing an error in console:

image

As mentioned, to show the flow just create a single line in the table for each e.g. canvas. So if across say 2 innovation packs there are 2 + 3 = 5 canvas templates then show 5 lines, with for now a simple text of "<InnovationPackName> - <canvasTemplateName>" + clicking it does the import. Hopefully somewhat clear - in parallel the end design will be worked so not worth spending effort making this look nice - about showing the functional flow...

techsmyth:
Really super to have this flow functional :)

Will work with Simone to get the polished design through...but already the following is a nice win for those helping to populate hubs:

image

valentinyanakiev (approved):

PR #2886, opened 2022-11-06
client updated to work with CardProfile on Aspect

techsmyth:
graphql queries / mutations are updated and codegen is running

Code is partially updated to work with new setup.


alstojanovic:
@techsmyth I fixed the build but there are issues in the server. Getting an 'Aspect not initialised' errors for all aspects.

alstojanovic (approved):

PR #3027, opened 2022-11-30
Update translation.en.json

larssondenise:


techsmyth (approved):
Key is that the URLs so do not change...

PR #3076, opened 2022-12-12
Update translation.en.json

larssondenise:
improved activity log entries


larssondenise:
@techsmyth if you can check:
for 'card-created': card description as part of the activity description and displayname as part of the action (otherwise we can leave it as I now did)
for 'member-joined': if it is correct :) 
+ overall check, thanks!

techsmyth (approved):

techsmyth:
@larssondenise thanks for the updates. I made the changes to get the card description available. Note that the dutch + other translation files will also need to be updated with this latest update - need to think through how to track that so that this kind of update gets handled.

PR #3122, opened 2022-12-15
fixed type of arguments to be UUID for aspect, callout

techsmyth:
The types were wrongly set on the api to be UUID_NAMEID, but only ever UUIDs should have been used


valentinyanakiev (changes requested):
That's not a fix, that's straight-up regression. It's incorrect that `UUID_NAMEID` is not used. Look up `HubAspect` or `HubAspectProvider` queries and you will see that.
`NAMEID` is used in places where the `Card` or `Canvas` nameID is building the url.

image

techsmyth:
@valentinyanakiev  that is new to me. I would say that we need to remove that, but indeed it can be later

valentinyanakiev:
The issue with UUID_NAMEID will be addressed separately.

PR #3288, opened 2023-01-04
updated to work with new library access under platform

techsmyth:
Updated queries to work with new location for library entity under platform


alstojanovic (approved):
Tested the client and works fine with no errors in the Graphql network. Tested library feature on hub templates.

PR #3325, opened 2023-01-10
language bar always in english; fixed new opportunity label

techsmyth:


ccanos:
@techsmyth This only affects the topbar, there is a language menu in the footer since not long ago that still translates:
image

Is that fine?

hero101 (approved):

PR #3571, opened 2023-02-24
Remove bad permission check

alstojanovic:

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


hero101 (approved):

hero101 (approved):
The code now accepts authentication errors (if you have read access to a resource) as something normal and handled by the client.
This is a simpler solution than generating twice the number of queries to check if the user has sufficient read access, just to not generate the error.

This of course requires a wider discussion but the solution was given by me, so I agree with it.

PR #3624, opened 2023-03-07
updated graphql queries to not retrieve agent => credentials 

techsmyth:
The client should not be determining roles / authorization based on credentials. 

The code has the graphql queries updated. 

Notes: 
- get the membership label for a user based on results from Community.myMembershipStatus field (server branch same name)
- remove all usage of the following function: get-user-role-translation-key.ts


valentinyanakiev:
Merged 3619-2 branch instead of this PR

PR #3709, opened 2023-03-21
updated codegen queries / mutations to compile against server-2542

techsmyth:
Client updates to work with profiles final stage on server.

Status:
- codegen is running
- client code is otherwise not yet updated


Comoque1:
callout creation from the client fails:

Comoque1:
The issue above was server side and is fixed already.

PR #3825, opened 2023-04-11
Added search results disclaimer 

valentinyanakiev:


src/domain/platform/search/SearchResultSection.tsx
```diff
@@ -27,11 +28,14 @@ const SearchResultSection: FC<ResultSectionProps> = ({
   loading,
 }) => {
   const titleWithCount = useMemo(() => `${title} (${results.length})`, [title, results.length]);
-
+  const { t } = useTranslation();
+  let resultDisclaimer: string | undefined = undefined;
+  if (results.length >= 8) resultDisclaimer = t('pages.search.results-disclaimer');
```
alstojanovic, line 33: Should be > 8 I think.

alstojanovic: const resultDisclaimer = results.length > 8 ? t('pages.search.results-disclaimer') : undefined

alstojanovic (changes requested):
Change the condition please.

alstojanovic (approved):

PR #4155, opened 2023-06-03
initial adjustment of graphql files to work with updated rooms api

techsmyth:
Updated graphql files for working with the new rooms api


src/domain/communication/discussion/providers/forum-queries.graphql
```diff
@@ -46,6 +46,14 @@ query platformDiscussions {
         comments {
           id
           messagesCount
+          messages {
```
techsmyth, line 49: This looks dangerous, as it will result in loading all messages from all rooms on the forum I think?

alstojanovic: Yes, this was previously commentsCount but now it became Comments. I am adjusting the Room interface on the client side to have messages optionally.

PR #4240, opened 2023-06-16
conditional logo button

hero101:
Created a special component just for the main logo. The reusable component was too complex to fix.


src/common/components/composite/layout/TopBar/LogoComponent.tsx
```diff
@@ -26,3 +27,16 @@ const LogoComponent = ({ url = DEFAULT_URL, ...rest }: LogoComponentProps) => {
 };
 
 export default LogoComponent;
+
+export const AlkemioLogoComponent = (props: BoxProps) => {
+  const { data } = useInnovationHubQuery();
```
ccanos, line 32: can we use `useUrlParams` here instead of making a query?

ccanos (changes requested):

ccanos (approved):

PR #4286, opened 2023-06-22
Innovation Hubs basic functionality

ccanos:
Goes together with a server PR: https://github.com/alkem-io/server/pull/2945

- Adds basic functionality for InnovationHub creation, see client-web/4091
- Solves a couple of comments from Simone, translations and little glitches
- Adds Referrer information to the 404 logging to Sentry.


src/services/logging/sentry/log.ts
```diff
@@ -38,6 +38,7 @@ export const info = (message: string, severity: typeof debugLevel | typeof infoL
 export const log404NotFound = () => {
   Sentry.withScope(scope => {
     scope.setLevel(errorLevel);
-    Sentry.captureEvent({ message: `404: ${document.location.href}` });
+    const message = `404: '${document.location.href}'` + (document.referrer ? ` Referrer: '${document.referrer}'` : '');
```
ccanos, line 41: Not sure if there is a better way to add an extra field to the sentry event and send together the url and the referer if there's any

hero101 (approved):

PR #4470, opened 2023-08-01
Sample sources

valentinyanakiev:
- List of sources
- Simplest solution out there
- Only supported way to present links with the Widget (out of the box)

The widget can have a custom embedded component as well. Sources are quite horribly returned in a freeform string, although it looks like the return type can actually be typed and could potentially return an array of results (which will require changes in the genai service, the server and the client). I couldn't find a better, quick solution.


techsmyth (commented):
Super links are showing up, that is the key thing. I think we do not need to show the second source details, just the links. So please remove line 28 and then merge this in.

PR #4510, opened 2023-08-08
Added privilege check for InnovationFlow Settings Dialog

valentinyanakiev:
There are multiple considerations here, hence the draft:
- We need to be provided with a good UX for web / mobile how this should look like so we don't re-iterate over it 5 times. It looks quite bad when the user doesn't have privileges, when the classification tagsets are long, when they are many etc... It needs proper thinking
- I am checking `UPDATE` privilege on the InnovationFlow State. However, this might not be the best course of action, as the settings dialog manages both the metadata / profile for the InnovationFlow (in that case, the privilege is the correct one) but it also can alter / update `Classification Tagsets` that are actually bound to `Callouts` and not the `Innovation Flow`. One option would be to check whether we can update the `Challenge` / `Opportunity`, but that also seems incorrect. @techsmyth  - any guidance here?

### DIFF
--- a/src/core/apollo/generated/apollo-hooks.ts
+++ b/src/core/apollo/generated/apollo-hooks.ts
@@ -8406,6 +8406,10 @@ export const ChallengeInnovationFlowStatesAllowedValuesDocument = gql`
             id
             state
           }
+          authorization {
+            id
+            myPrivileges
+          }
           profile {
             id
             tagsets {
@@ -8490,6 +8494,10 @@ export const OpportunityInnovationFlowStatesAllowedValuesDocument = gql`
           lifecycle {
             id
             state
+          }
+          authorization {
+            id
+            myPrivileges
           }
           profile {
             id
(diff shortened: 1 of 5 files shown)


techsmyth:
@valentinyanakiev  there are few items being raised which I think need to be decoupled from each other. 

I have created a new issue for the UX feedback #4516 

The checking of the UPDATE privilege on the InnovationFlow entity is fine for whether to show this control or not. In terms of being able to modify the set of flow states allowed via the TagsetTemplate being used for that, that is also fine - as the InnovationFlow is logically where those values come from. So if you can update the description etc then you should also be able to modify the set of flow states. 

So this one is good to go logically if it runs functionally.

valentinyanakiev:
@techsmyth Well, permission-wise, this is a bit more complicated, unfortunately. Yes, the allowedValues and the current state value are synchronized when the InnovationFlow's lifecycle changes. But that happens in the services without any auth checks for the tagsets. If the validation passes on API level, and the change requires an admin, then it's all considered fine. 
The problem is, here we have 2 decoupled entities (or a set of entities, in one case).
- The `InnovationFlow` itself - which is OK to be updated with `Update` on it. There can be no other way.
- The `Classification Tagsets` - they are actually bound to the `Callouts`, not to the `Innovation Flow`. If we have `N` `Callouts`, we have `N+1` tagsets, the `1` coming from the `Innovation Flow` one. Currently, it controls the `allowedValues` and the default value. The problem is, the rest of the `Tagsets` are coming from the `Profile` of the `Callouts`. And this is, logically, not tied in any way to the `Innovation Flow`. So, in theory, having permissions to update the `Innovation Flow` does not mandate having permission to update a `Tagset` bound to a `Callout`. Unless, of course, we check a `role`, not `privilege` - but we have moved away from that.

PR #4524, opened 2023-08-10
Added privilege check for updating innovationFlow template

valentinyanakiev:


src/domain/platform/admin/templates/InnovationTemplates/UpdateInnovationFlow.tsx
```diff
@@ -37,6 +38,15 @@ const UpdateInnovationFlow: FC<EditLifecycleProps> = ({
...
     template => lifecycle && lifecycle.templateName && template.definition.includes(lifecycle.templateName)
   );
+  const { data } = useInnovationFlowAuthorizationQuery({
+    variables: { innovationFlowId: entityId! },
+    skip: !entityId,
+  });
+  const myPrivileges = data?.lookup.innovationFlow?.authorization?.myPrivileges;
+
+  const privileges = {
+    canUpdate: (myPrivileges ?? []).some(p => p === AuthorizationPrivilege.UpdateInnovationFlow),
```
ccanos, line 48: I know it's everywhere, probably you copy pasted this `some` from somewhere, but this can be simplified to `includes(AuthorizationPrivilege.UpdateInnovationFlow)` and we are tending to use that more

ccanos (changes requested):
super small change requested

PR #4641, opened 2023-08-28
Fix orientation type

valentinyanakiev:


hero101 (changes requested):
Fix quotes.
Something is wrong with the commit history.

PR #4863, opened 2023-09-29
Fix debug ids

valentinyanakiev:
- had to revert the auto-commit feature as it was failing
- this should be working as a setup, deployed on `acc` for validation

if this is OK we can deploy it on prod on Monday so we start tracking the issues properly


vite.sentry.config.mjs
```diff
@@ -35,9 +35,6 @@ export default defineConfig({
         name: `client-web@v${version}`,
         version,
         create: process.env.NODE_ENV === 'production',
-        setCommits: {
```
hero101, line 38: why was it removed?

hero101 (changes requested):

hero101 (approved):

PR #4864, opened 2023-10-01
ran codegen after server updates

techsmyth:


valentinyanakiev:
Duplicated, fixed with another PR.

PR #4938, opened 2023-10-11
Updated worfklows to use latest actions 

valentinyanakiev:
- used latest version of all actions
- updated deprecated workflow methods


.github/workflows/build-deploy-k8s-dev-azure.yml
```diff
@@ -33,12 +33,12 @@ jobs:
           echo ::set-output name=created::$(date -u +'%Y-%m-%dT%H:%M:%SZ')
```
hero101, line 34: there are two `set-output` on line 32,33

hero101 (changes requested):

hero101 (approved):

PR #4957, opened 2023-10-16
StorageBucket --> StorageAggregator

valentinyanakiev:


techsmyth (dismissed):
Changes make sense and are running, gets the client compiling and then we pick up using the full api later.

PR #5056, opened 2023-10-29
Usage of License entity; Setting of feature flags on spaces

techsmyth:
Working with Space Visibility now being part of License entity.

Allow setting of feature flags on Space via platform settings

Done:
- Space platform settings to be tidied up e.g. to have the current feature flag value + generally code tidy up
- Validate usage of new Privileges for working with WhiteboardRt and Callout Template creation from callout.
- CalloutTypeSelect to use Collaboration privileges, not platform privileges to determine what callout types can be created

Todo:
- Disable WhiteboardRts when they are already created, (gray overlay), is checking CreateWhiteboardRt permission on the callout. That doesn't fully make sense?
- Actual review.


ccanos:
@techsmyth I have pushed all my changes. I have updated the description and will wrap the remaining things up tomorrow morning. I need to talk with you about the gray overlay.

PR #5095, opened 2023-11-01
License entity, plus mgmt

techsmyth:
Working with Space Visibility now being part of License entity.

Allow setting of feature flags on Space via platform settings

Done:
- Space platform settings to be tidied up e.g. to have the current feature flag value + generally code tidy up
- Validate usage of new Privileges for working with WhiteboardRt and Callout Template creation from callout.
- CalloutTypeSelect to use Collaboration privileges, not platform privileges to determine what callout types can be created
- Disable WhiteboardRts when they are already created, (gray overlay), is checking CreateWhiteboardRt permission on the callout. That doesn't fully make sense?


src/domain/journey/space/license/LicenseFeature.ts
```diff
@@ -0,0 +1,5 @@
+// TODO: Export these from the server
```
techsmyth, line 1: There is now LicenseFeatureFlagName enum exposed on the server.

ccanos (approved):

PR #5251, opened 2023-11-22
Release notes dialog only opening on new entry

hero101:
The translation file now holds a list with entries for different release notes.
The latest must be on top, with a key with a higher value than the ones before it - **see image below**.
Every new entry on top of the list, with a properly set key will be displayed to the user.
As long as there is no new entry, the dialog won't be shown again - **notes are now not tided to app version.**

image

---
> **Example**: The next entry in the list must be added on top and with a key "3".
image


hero101:
@ccanos Converted to draft - I need to change the translation file structure to avoid overrides for new values.

ccanos (dismissed):

ccanos (dismissed):

PR #5356, opened 2023-12-22
 MyDashboard: Platform release notices 2

ccanos:
This PR is a continuation of #5307 which was closed because I saw this other branch `client-5010-2` that had an update over `client-5010`

Requires some work fixing the conflicts, or picking up the changes and applying them to the current develop.


ccanos:
I don't think this is needed anymore

PR #5674, opened 2024-03-03
Innovation flow: switch to states, simplified api usage

techsmyth:
also simplifies api usage for working with innovation flows on challenge / opportunity

Key changes:
- move to looking up by uuid instead of nameids for innovation flows
- add mutation in for updating the template in usage on challenge / opportunity: search on useUpdateInnovationFlowLifecycleTemplateMutation
- New simple display for lifecycle states
- client to be updated to use new mutations / setup, including being able to manage the set of states + select the state that should be active


techsmyth:
@ccanos there is an error when trying to add an innovation flow template to an innovation pack: 
image
can you please have a look?

ccanos (approved):

ccanos (approved):

PR #5729, opened 2024-03-15
DisplayLocation ==> CalloutGroup

techsmyth:
updated to work with new management of groups from the server


Comoque1:
The server is not starting
image

PR #5780, opened 2024-03-22
Hide calendar if no events and no admin

EmiliaPavlova:
Hiding Events block when there are no events if the viewer is no admin or a lead. If they are, showing the updated message.

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -82,7 +82,7 @@
     "past-events": "Past Events",
     "show-calendar": "Show calendar",
     "defaultEventDescription": "$t(common.location):\n\n$t(common.url):\n\n$t(common.description):",
-    "admins-only": "Since there are no events, this block is only visible for admins and leads"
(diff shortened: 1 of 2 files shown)


src/core/i18n/en/translation.en.json
```diff
@@ -81,7 +81,8 @@
     "no-upcoming-events": "Currently, there are no upcoming events to be displayed",
     "past-events": "Past Events",
     "show-calendar": "Show calendar",
-    "defaultEventDescription": "$t(common.location):\n\n$t(common.url):\n\n$t(common.description):"
+    "defaultEventDescription": "$t(common.location):\n\n$t(common.url):\n\n$t(common.description):",
+    "admins-only": "Since there are no events, this block is only visible for admins and leads"
```
ccanos, line 85: please use camelCase for new translations

ccanos (changes requested):
code looks good, just a style comment

ccanos (approved):

Issue #5750, opened 2024-03-19
Move "Recent Activity/Most Active" block to be dialog


SimoneZaza:
## Description

As a user, I don't get energized when the first thing I see when I open a space is that the last activity happened 3 months ago. 
As an admin, I want to use this in-your-face spot for my most important content in this space, or for a guiding post, not for the most recent activity
As a user, I want to be able to see the most recent activity from any page on the Space, keeping my flow when I quickly check for updates. 

## Acceptance criteria

- [ ] Dialog for recent activities and most active as per design
- [ ] Icon in tabmenu to open this dialog
- [ ] Top of the right column of the Dashboard/Home of the (sub)Space page to be no longer split in 2 columns, but to just be one (see design)
- [ ] Mobile should also be added to mobile menu

## Additional Context

Figma: https://www.figma.com/file/XDWND6E1djfELwBfMa4U3u/Spaces-%26-SubSpaces?type=design&node-id=7%3A3889&mode=design&t=JuCyfWWG1LR3Bb5E-1

## Areas that will be affected 
 
_To be added during the refinement_

PR #5974, opened 2024-04-22
Activity dialog for subspaces

bobbykolev:
### Description

The two activity lists were merged into a dialog (subspace level) as per request: https://github.com/alkem-io/client-web/issues/5750

The current data fetch is not ideal but fixing it should be part of a separate task (comments left).

### DIFF
--- a/src/domain/journey/common/Activity/ActivityDialog.tsx
+++ b/src/domain/journey/common/Activity/ActivityDialog.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import { DialogContent, useMediaQuery } from '@mui/material';
+import { Theme } from '@mui/material/styles';
 import DialogWithGrid from '../../../../core/ui/dialog/DialogWithGrid';
 import DialogHeader from '../../../../core/ui/dialog/DialogHeader';
-import { DialogContent } from '@mui/material';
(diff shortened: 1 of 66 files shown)


src/domain/journey/common/Activity/ActivityDialog.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useRouteResolver } from '../../../../main/routing/resolvers/RouteResolver';
+import DashboardAllContributionsBlock from '../dashboardRecentContributionsBlock/DashboardAllContributionsBlock';
+
+export interface SubspacesListDialogProps {
+  open?: boolean;
+  onClose?: () => void;
+  journeyId: string;
+}
+
+const SubspacesListDialog = ({ open = false, journeyId, onClose }: SubspacesListDialogProps) => {
```
ccanos, line 17: `ActivityDialog` ?

ccanos (dismissed):
just a small comment, the rest looks fine to me

src/domain/journey/common/Activity/ActivityDialog.tsx
bobbykolev: thanks for spotting it 🙏

bobbykolev:
> Please have a look at the comments

Sure, thanks for looking at it 🙏

PR #5980, opened 2024-04-22
New Create Collaboration tool button for Subspace

EmiliaPavlova:
- Add vertical separator + new "+ Collaboration Tool" button to the right side of the innovation flow
- In mobile, keep the button as is
- Remove large add button from the subspaces

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -647,6 +647,7 @@
       "description": "We apologize, but the content you requested is not available. It may have been deleted, or you may not have the permission to access it, such as if it was put in draft mode."
     },
     "createFull": "Add collaboration tool",
+    "createShort": "Collaboration tool",
     "contributions": "{{count}} Contribution",
     "contributions_plural": "{{count}} Contributions",
     "draftNotice": "DRAFT - Only visible for admins",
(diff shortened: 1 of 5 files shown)


src/core/i18n/en/translation.en.json
```diff
@@ -647,6 +647,7 @@
       "description": "We apologize, but the content you requested is not available. It may have been deleted, or you may not have the permission to access it, such as if it was put in draft mode."
     },
     "createFull": "Add collaboration tool",
+    "createShort": "Collaboration tool",
```
ccanos, line 650: I think I'd do `"$t(common.collaborationTool)"` and add `collaborationTool` there in `common`, I've been about to add it already a few times.

src/domain/collaboration/callout/CalloutsInContext/CalloutsGroupView.tsx
```diff
@@ -13,6 +13,7 @@ interface CalloutsGroupProps extends CalloutsViewProps {
   groupName: CalloutGroupName;
   flowState?: string;
   createButtonPlace?: 'top' | 'bottom';
+  hideButton?: boolean;
```
ccanos, line 17: can we call this `hideCreateButton`? 
I would even try to use `createButtonPlace` if it's clean. something like `'top' | 'bottom' | 'none'`

ccanos (changes requested):
Just a couple of aesthetic comments

PR #6009, opened 2024-04-25
Add navigation for subspace settings

EmiliaPavlova:


src/domain/journey/subspace/routing/SubspaceRoute.tsx
```diff
@@ -27,13 +28,18 @@ const SubspaceRoute = () => {
...
         <Route path={SubspaceDialog.Activity} element={<SubspaceHomePage dialog={SubspaceDialog.Activity} />} />
         <Route path={SubspaceDialog.Timeline} element={<SubspaceHomePage dialog={SubspaceDialog.Timeline} />} />
         <Route path={SubspaceDialog.Share} element={<SubspaceHomePage dialog={SubspaceDialog.Share} />} />
         <Route
           path={`${SubspaceDialog.Timeline}/:${nameOfUrl.calendarEventNameId}`}
           element={<SubspaceHomePage dialog={SubspaceDialog.Timeline} />}
         />
+        <Route
+          path={SubspaceDialog.Settings}
+          element={journeyId && <Navigate to={`${buildJourneyAdminUrl(journeyId)}/profile`} />}
```
ccanos, line 41: don't use the `journeyId` here.
This is going to work, but it's going to generate a url like this:
image
which works just because on the server it doesn't matter if we query `id` or `nameId`.

I'm taking a look myself, please don't merge this

ccanos (changes requested):

ccanos:
@EmiliaPavlova I have changed a bit your approach. I think this is cleaner but please let @me-andre take a look, maybe he has a better idea

PR #6068, opened 2024-04-29
Show loader if data is still loading for About screen

EmiliaPavlova:
Add loader for About page when spaceId is not yet present.


bobbykolev:
The correct place to check and show the `Loading` is in the `JourneyAboutDialog` where you have access to the `loading` prop from the AboutPageContainer (doing the data queries).

bobbykolev (approved):

PR #6078, opened 2024-04-29
Fixes

valentinyanakiev:
- collaboration should not be queried on space / subspace cards


bobbykolev:
We're using collaboration.id in the innovation flow and the current changes would break the current subspace implementation.
I'll assign @me-andre to double-check and suggest a path for adopting the changes.

PR #6118, opened 2024-05-04
Global roles

techsmyth:
Renamed global roles

Retrieve but not yet use new setting on space setting for privacy

Mutations for platform roles combined into one with the role being specified.
Mutations for organizaton roles combined into one with the role being specified

Todo: simplify code that is now fully duplicated.


ccanos (dismissed):

package-lock.json
bobbykolev, line 1: The package-lock changes seem unintentional. Let's discard them.

bobbykolev:
Looking good! We could improve the repetition of Tabs in `PlatformRoleAssignementPage` but this is more of a nice-to-have. Green light from me, only the package-lock change could be discarded.

package-lock.json
techsmyth: ok, can you please do the update? I have otherwise merged in from develop...

bobbykolev (approved):

ccanos (dismissed):

ccanos (approved):

PR #6159, opened 2024-05-09
Space administration goes under space.url/settings

me-andre:
### What is left:

Wrap Space/Subspace administration route components into proper checks.


src/domain/journey/subspace/subspaceHome/SubspaceHomePage.tsx
```diff
@@ -158,7 +157,6 @@ const SubspaceHomePage = ({ dialog }: SubspaceHomePageProps) => {
                     dialogType={SubspaceDialog.Settings}
                     label={t(`spaceDialog.${SubspaceDialog.Settings}` as const)}
```
bobbykolev, line 158: It's not directly related to this PR but using the enum values from SubspaceDialog for translations is not ideal. I think that using the string as 'spaceDialog.settings' is better (for all translations). If we were using the  SubspaceDialog enum to build the translation strings, then it would make sense.

bobbykolev (commented):
Awesome change @me-andre 
Now we have a clear separation between the GA routing and Space/SubSpace one allowing better handling of different roles and privileges. 👏

Should we change the restriction in `NonAdminRedirects` form `AuthorizationPrivilege.Admin` to `AuthorizationPrivilege.PlatformAdmin`?

Besides restricting the storage tabs, and the community access on an opportunity level, everything looks and works fine.

bobbykolev (approved):

PR #6215, opened 2024-05-20
Draft innovation hub with account

valentinyanakiev:
- We need a `displayName` for the `account`
- Not all `innovationHubs` have an `account`, so `innovationHubs` query with a property selection of `account` throws an exception
- if we omit the `account` from the query, we won't refetch the values correctly...

### DIFF
--- a/src/core/apollo/generated/apollo-helpers.ts
+++ b/src/core/apollo/generated/apollo-helpers.ts
@@ -5,6 +5,7 @@ export type APMFieldPolicy = {
   rumEnabled?: FieldPolicy<any> | FieldReadFunction<any>;
 };
 export type AccountKeySpecifier = (
+  | 'agent'
   | 'authorization'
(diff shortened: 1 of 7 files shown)


techsmyth:
@valentinyanakiev why don't we change the create logic? Then we create the InnovationHub from within the Account context.  That is also logically how it should be...the global creation of an innovation hub / innovation pack / VC just goes away. The only thing that remains there is then the option for re-assigning between accounts - which is a very special operation.

techsmyth:
@valentinyanakiev lets put this out a bit as discussed, the server-3900 PR makes the account field optional so this PR can be done later.

valentinyanakiev:
@techsmyth closing this as it is not needed now. Will fix it once we have the `Account` with `Innovation Hub`, `Innovation Pack`, `Virtual Contributor` and a proper design.

Issue #6247, opened 2024-05-27
BUG: Invitations records are not displayed on subspace admin section


Comoque1:
**Describe the bug**

**To Reproduce**

1. Go to http://localhost:3000/111/challenges/challengeinvitations/settings/community (or subspace admin/community)
2. Make an invitation - it is not displayed
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNjhSQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e7889060dbc326132e18fac702813f4a69e172c9/image.png" alt="image.png" />

**Expected behavior**
Invitations on subspace level are displayed the same way as on space level.

Comoque1:
Works fine now.

PR #6249, opened 2024-05-27
Add invitations on subspace community admin level

bobbykolev:
https://github.com/alkem-io/client-web/issues/6247

### DIFF
--- a/src/domain/journey/subspace/pages/AdminSubspaceCommunityPage.tsx
+++ b/src/domain/journey/subspace/pages/AdminSubspaceCommunityPage.tsx
@@ -33,9 +33,13 @@ const AdminSubspaceCommunityPage: FC<SettingsPageProps> = ({ routePrefix = '../'
     virtualContributors,
     applications,
     invitations,
+    invitationsExternal,
     communityPolicy,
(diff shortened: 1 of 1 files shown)


src/domain/journey/subspace/pages/AdminSubspaceCommunityPage.tsx
```diff
@@ -83,6 +87,12 @@ const AdminSubspaceCommunityPage: FC<SettingsPageProps> = ({ routePrefix = '../'
             <CommunityApplications
               applications={applications}
               onApplicationStateChange={onApplicationStateChange}
+              canHandleInvitations
```
EmiliaPavlova, line 90: I don't think you need this prop.

EmiliaPavlova (approved):

src/domain/journey/subspace/pages/AdminSubspaceCommunityPage.tsx
bobbykolev: It's required for the header of the block. Without it it says only Applications. When provided (value true) it states ' & invitations'

Issue #5911, opened 2024-04-15
Improved editing of VC profiles


techsmyth:
## Description

As a sales lead I want the VCs to have rich profiles

## Acceptance criteria

- [ ] Allow editing of all content on a VC profile, using user profile editing page as a basis:
- Avatar
- Name
- Tagline
- Description
- Tags
- References
- Host (can't edit)
- Space (can't edit)
- Subspace (dropdown with list of subspaces within the related space)

## Additional Context
Figma: https://www.figma.com/design/0cSP9R1SqhlrBuucR1NYJu/Create-Virtual-Contributor?node-id=194-1879&t=9psLVx6UF7uJzYIS-1

- Who can edit? Admin or only host?
- Complexity with space relation + goign back and forth between teams

SimoneZaza:
@techsmyth who has the privilige to edit the profile/create the vc? Only the host or also the admin?

bobbykolev:
I'll extract two points from the inital ACs:
- references;
- Dropdown with Subspaces - MVP shows only the selected one. You can change the BoK by deleting the current VC and creating a new one.
@SimoneZaza @Comoque1 FYI

Comoque1:
thanks @bobbykolev !

PR #6273, opened 2024-05-30
VC Edit Profile 

bobbykolev:
https://github.com/alkem-io/client-web/issues/5911

### DIFF
--- a/src/domain/community/virtualContributor/vcSettingsPage/VCSettingsPage.tsx
+++ b/src/domain/community/virtualContributor/vcSettingsPage/VCSettingsPage.tsx
@@ -10,9 +10,13 @@ import VirtualContributorForm from './VirtualContributorForm';
 import PageContentColumn from '../../../../core/ui/content/PageContentColumn';
 import PageContentBlock from '../../../../core/ui/content/PageContentBlock';
 import PageContent from '../../../../core/ui/content/PageContent';
+import { useNotification } from '../../../../core/ui/notifications/useNotification';
+import { StorageConfigContextProvider } from '../../../storage/StorageBucket/StorageConfigContext';
 
 export const VCSettingsPage = () => {
   const { vcNameId = '' } = useUrlParams();
+
+  const notify = useNotification();
 
   const { data, loading } = useVirtualContributorQuery({
     variables: {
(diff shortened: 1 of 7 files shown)


src/domain/community/virtualContributor/vcSettingsPage/VCSettingsPage.tsx
```diff
@@ -13,11 +17,31 @@ export const VCSettingsPage = () => {
     },
   });
 
+  const [updateContributorMutation] = useUpdateVirtualContributorMutation();
+
+  const handleUpdate = async virtualContributor => {
```
ccanos, line 22: I would just call the function here, there's no need to await if you're not doing anything after

src/domain/community/virtualContributor/vcSettingsPage/VCSettingsPage.tsx
```diff
@@ -13,11 +17,31 @@ export const VCSettingsPage = () => {
...
+          ID: virtualContributor.ID,
+          profileData: virtualContributor.profileData,
+        },
+      },
+    });
+  };
+
   if (loading) return <Loading text={'Loading Virtual Contributor Settings ...'} />;
 
+  // TODO: StorageProvider for the VC
```
ccanos, line 35: Normally we just use the platform StorageProvider for entities that we are creating (which is definitely wrong, but it is a solution if you need it now), 
or hide the ReferencesSegment in the Create form and show it in the edit form. Which is also bad but I think it's better better.

src/domain/community/virtualContributor/vcSettingsPage/VirtualContributorForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+    name: displayName,
+    nameID: nameID,
+    description: description ?? '',
+    tagline: tagline,
+    avatar: avatar,
+    tagsets: tagsets,
+  };
+
+  const validationSchema = yup.object().shape({
+    name: nameSegmentSchema.fields?.name || yup.string(),
```
ccanos, line 75: probably copied from somewhere else but please use `??` for nullish coalescing

src/domain/community/virtualContributor/vcSettingsPage/VirtualContributorForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  const backButton = (
+    <Grid item>
+      <WrapperButton variant="default" onClick={handleBack} text={t('buttons.back')} />
+    </Grid>
+  );
+
+  const HostFields = () => (
+    <>
+      <FormikInputField name={'host'} title={'Host'} required readOnly disabled />
```
ccanos, line 122: please use double quotes here instead of `{'host'}` ... 
All if this is temporary, right? can you add a comment?

src/domain/community/virtualContributor/vcSettingsPage/VirtualContributorForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+      const originalTagset = tagsets?.find(value => value.name === updatedTagset.name);
+      if (originalTagset) result.push({ ...originalTagset, tags: updatedTagset.tags });
+    });
+
+    return result;
+  };
+
+  const [handleSubmit, loading] = useLoadingState(async (values: VirtualContributorFromProps) => {
+    const { tagsets, description, tagline, name, ...otherData } = values;
+    const updatedTagsets = getUpdatedTagsets(tagsets || []);
```
ccanos, line 92: `||` => `??` same here

src/domain/community/virtualContributor/vcSettingsPage/VirtualContributorForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    const virtualContributor = {
+      ID: currentVirtualContributor.id,
+      profileData: {
+        displayName: name,
+        description,
+        tagline,
+        tagsets: updatedTagsets.map(r => ({
+          ID: r.id,
```
ccanos, line 101: is this `id: undefined` required? you're creating a new object...
also I would call the variable `tagset`

src/domain/community/virtualContributor/vcSettingsPage/VirtualContributorForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+    name: nameSegmentSchema.fields?.name || yup.string(),
+    nameID: nameSegmentSchema.fields?.nameID || yup.string(),
+    description: profileSegmentSchema.fields?.description || yup.string(),
+  });
+
+  const getUpdatedTagsets = (updatedTagsets: Tagset[]) => {
+    const result: UpdateTagset[] = [];
+    updatedTagsets.forEach(updatedTagset => {
+      const originalTagset = tagsets?.find(value => value.name === updatedTagset.name);
+      if (originalTagset) result.push({ ...originalTagset, tags: updatedTagset.tags });
```
ccanos, line 84: please use curly brackets here

ccanos (changes requested):
just a few style comments, but logic wise looks good to me

src/domain/community/virtualContributor/vcSettingsPage/VirtualContributorForm.tsx
bobbykolev: I'm displaying the tagsets, there's no default tagset. I could use some help on switching from default keywords and capabilities to just tags.

Issue #6437, opened 2024-06-21
VC Profile Settings page: tab to display space memberships, with option to leave


techsmyth:
## Description

As a provider of a VC, I want to be able to see, and potentially stop, membership of spaces that my VC is in

## Acceptance criteria
- [ ] VC Profile settings page to have a list of spaces the VC is a member of
- [ ] Allow leaving button
- [ ] Confirmation dialog for leaving
- [ ] Ability to see the settings if you have the "UPDATE" privilege
- [ ] Ability to leave a space membership requires the "GRANT" privilege

Note: do not show space memberships on the public profile for a VC.

General: mimic the user profile memberships management

TBD: unlink VC from Space mutation - existing but check the logic behind it (removeCommunityRoleForVirtual); 

## Additional Context
<img width="1482" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNldBQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--696cc8de7eea6e18dc0502d13d9e172efbfa73ec/image.png" alt="image.png" />

https://www.figma.com/design/WFcbewbnhPqYpGBzBJaPq1/VC-Profile-Settings?node-id=0-1&t=HPhsHqfBBhmA7pxq-1

PR #6472, opened 2024-06-26
VC Memberships page

ccanos:
#6437 
- Cleaned up the userProvider
- Added a context PendingMembershipsDialog, to be able to open the dialog from the top right popup menu and from the Pending memberships cards.
- Added the Memberships tab on VC settings


techsmyth:
@ccanos super clean up. I cannot run it now / do a detailed review, so will leave that for Bobby. It will be interesting to see when this is included in a release what the userProvider response time becomes on production...at least for me it is on the order of 200ms, but then again I do have a lot of credentials.

src/domain/community/user/pages/UserMembershipPage.tsx
```diff
@@ -8,36 +8,84 @@ import UserSettingsLayout from '../../../platform/admin/user/layout/UserSettings
...
+        contributorId: userMetadata?.user.id!,
+        contributorType: CommunityContributorType.User,
+      };
+      acc.push(currentSpace);
+
+      const subspaces = space.subspaces.map(subspace => ({
+        id: subspace.id,
+        spaceID: subspace.id,
+        spaceLevel:
+          subspace.type === SpaceType.Challenge
```
bobbykolev, line 52: Can we query directly the level instead of the type and skip this obsolete logic?

bobbykolev (approved):
Looking good, @ccanos 
Only the pending invitations section needs an action, but it could be used only for visibility purposes.

bobbykolev (approved):

PR #6485, opened 2024-06-27
Filter only the public VCs from the Platform Library

bobbykolev:
https://github.com/alkem-io/client-web/issues/6484


Comoque1 (approved):

Comoque1:
Tested OK locally.

PR #6497, opened 2024-06-28
updated so that memberJoined works with contributor

techsmyth:
To work with server branch that updates the type


valentinyanakiev (approved):
Will test on acc

Issue #6476, opened 2024-06-27
BUG: platform invitations are processed before user is registered


techsmyth:
**Describe the bug**
A user that signs up to the platform after being invited by email is triggering an error in the server logs.

I think that this is due to communityInvitations being triggered before the user is registered:
<img width="1325" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOXVXQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3d06d76bc23002dc77a5108626c9f82cea797446/image.png" alt="image.png" />

resulting in:
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOXlXQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c44c9fa344292bedf5328d70c14c3a784443d654/image.png" alt="image.png" />

and 

<img width="888" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK0dXQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8a0ce3148292a2ef85441d413f28bd9d89519a55/image.png" alt="image.png" />

The result is that the user does not see their pending invitations when they are logged in. 

**To Reproduce**

1. Invite an external user by email
2. Login with that user so that registrations are tiggered
3. Error is generated in the logs

**Expected behavior**
The user registration flow should complete before communityInvitations are retrieved

PR #6504, opened 2024-06-28
Fix user getting error messages in the dashboard on the first login

ccanos:
Fix bug #6476 
Users were redirected to the dashboard after login, and on the first login they don't have yet their profile created.
That was causing a lot of errors in other queries.
UserProvider (useUserContext), creates the profile the first time it's called if it doesn't exist yet. I think we need to refactor that in the future.
For now I have added a call to useUserContext to the loginSuccessPage.
That will make the login slower but at least we make sure the profile is fully created when the user is redirected.

We could also trigger a smaller request to from the login page and create the user there. I'll check on monday because I think we had a task about refactoring that profile creation, but I haven't seen it, probably @techsmyth remembers

also added two skips to `usePendingMemberships`, which I think were causing issues too


bobbykolev (approved):
The fix makes sense. However, these changes require retesting of all the login/register flows.

PR #6525, opened 2024-07-03
Redirecting to Restricted in case of unauthorized access to a VC

me-andre:
As this includes update to Vite 5, please don't merge before the upcoming release.


bobbykolev (approved):
The RestrictedRoute logic looks and works good already blocking the access to external VCs from regular users.

On the Vite changes, please make sure there are no conflicting dependencies when running npm install. It would be ideal to be in a separate PR.

Issue #6480, opened 2024-06-27
VC community invite with URL


bobbykolev:
## Description

As an Admin user, I want to be able to invite a VC to my community via its profile URL.

## Acceptance criteria

- [ ] Allow the user to enter the URL to the profile of a VC that they want to invite. 
- [ ] The same invite flow should be presented as the external one - invite from Platform Library.
- [ ] Allow the user to click the plus button to open the invite dialog where they can add a message.

## Additional Context
Extracted from:
https://github.com/alkem-io/client-web/issues/6317

Screen shots / UX designs / data that are relevant.
Note that insead introducing new search/result, we're reusing the current search and results table.
https://www.figma.com/design/oKRFyc7P9WwlNyIWAiiHKf/Inviting?node-id=76-284&t=FMEAi1gdenS9hEn9-1

bobbykolev:
With this implementation it would be trevial to add URL support for adding memebers to community (user & Orgs).

Comoque1:
Tested OK on DEV env.
@techsmyth @SimonDoherty1234 I am closing this story, as the AC is met, but with the current server/client implementation only **public and listed** vc can be invited. For hidden it is not the case.

If it should be extended, we can raise another story.

PR #6527, opened 2024-07-03
Allow searching for external VCs by URL

bobbykolev:
https://github.com/alkem-io/client-web/issues/6480

The same Search was reused instead of duplicating both the search and result/s table (the story will be updated).

The `CommunityAddMembersDialog` now supports search of entities by URL, just by passing a flag, it could be easily enabled for users/orgs.

### DIFF
--- a/src/domain/community/community/CommunityAdmin/CommunityAddMembersDialog.tsx
+++ b/src/domain/community/community/CommunityAdmin/CommunityAddMembersDialog.tsx
@@ -10,8 +10,10 @@ import { Identifiable } from '../../../../core/utils/Identifiable';
 import Gutters from '../../../../core/ui/grid/Gutters';
 import DialogHeader from '../../../../core/ui/dialog/DialogHeader';
 import LoadingIconButton from '../../../../core/ui/button/LoadingIconButton';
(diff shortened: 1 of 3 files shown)


src/domain/community/community/CommunityAdmin/CommunityAddMembersDialog.tsx
```diff
@@ -56,6 +64,23 @@ const CommunityAddMembersDialog: FC<CommunityAddMembersDialogProps> = ({ onClose
...
+    let filterValue = event.target.value;
+
+    const urlNameId = getURLPath(filterValue)?.split('/')[2];
+
+    // in case of a URL value split the pathname, extract the nameID, map it to displayName and apply filter
+    if (allowSearchByURL && urlNameId) {
+      const allEntities = await fetchAvailableEntities?.();
+
+      filterValue =
+        allEntities?.filter(entity => entity.nameID === urlNameId).map(entity => entity.profile.displayName)?.[0] ||
```
ccanos, line 77: Here you're coallescing, please use ??

ccanos (approved):
Just one small comment you can ignore. I havent tested it though, will do it later, may do it on dev

PR #6551, opened 2024-07-09
Add VC blocks in Space Community and Subspace About and Contributors

EmiliaPavlova:
-  Block on community tab in space
-  Block on contributors dialog in Subspace
-  Listing a max of 3 VCs to be shown, more when opening dialog
-  VC's visible under contributors block in About dialog
-  If no VCs then do not show the block
-  If VC is private / hidden, do not show in About dialog (possible?)
-  Only show the block if the user has READ access to the space / subspace


src/domain/journey/space/SpaceCommunityPage/SpaceCommunityPage.tsx
```diff
@@ -89,6 +100,20 @@ const SpaceCommunityPage = () => {
 
   const [isActivitiesDialogOpen, setIsActivitiesDialogOpen] = useState(false);
 
+  const { data: virtualContributorsData, loading } = useVirtualContributorsMySpacesQuery();
```
bobbykolev, line 103: I don't think you should gather the VCs from 'me->mySpaces'. It's logical to list the VC in the community similar to how we're gathering the other Space community members. 

It should be the contributors to that Space, not specifically part of 'your' spaces.

src/domain/journey/subspace/dialogs/ContributorsToggleDialog.tsx
```diff
@@ -45,12 +48,19 @@ const ContributorsToggleDialog = ({ open = false, journeyId, onClose }: Contribu
       contributorType: ContributorType.Organizations,
     }));
 
+  const virtualContributors: VirtualContributorProps[] =
+    data?.lookup.space?.account.virtualContributors.filter(vc => vc.searchVisibility === SearchVisibility.Public) ?? [];
```
bobbykolev, line 52: Hm, if we have external VCs then they would be part of the community but not part of its Account, right?

src/domain/community/community/VirtualContributorsBlock/VirtualContributorsBlock.tsx
```diff
@@ -0,0 +1,61 @@
...
+            to={vc.profile.url}
+          >
+            <BlockSectionTitle>{vc.profile.displayName}</BlockSectionTitle>
+          </BadgeCardView>
+        ))}
+        {virtualContributors.length > VIRTUAL_CONTRIBUTORS_LIMIT && (
+          <SeeMore label="buttons.see-more" onClick={openDialog} />
+        )}
+      </PageContentBlock>
+      <VirtualContributorsDialog open={dialogOpen} onClose={closeDialog} virtualContributors={virtualContributors} />
```
bobbykolev, line 56: You can add the dialog inside the PageContent and remove the fragment.

src/domain/journey/common/AboutPageContainer/AboutPageContainer.tsx
```diff
@@ -85,6 +89,12 @@ const AboutPageContainer: FC<AboutPageContainerProps> = ({ journeyId, children }
   });
 
   const memberProfile = membersData?.lookup.space?.profile;
+  const virtualContributors = membersData?.lookup.space?.account.virtualContributors.filter(
```
bobbykolev, line 92: Again, this would not return any external VCs part of the community.

src/domain/community/community/VirtualContributorsBlock/VirtualContributorsBlock.tsx
bobbykolev: redundant fragment, nothing critical

src/domain/journey/subspace/dialogs/ContributorsToggleDialog.tsx
EmiliaPavlova: Being on the subspace level, I cannot reach the VCs which are stored on space level and not on subspace level. So until there is an API endpoint to fetch all VCs in a space form a subspace or at least to get the spaceId, the only way to get them is in the account.

src/domain/community/community/VirtualContributorsBlock/VirtualContributorsDialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+    props: ListItemButtonProps<D, P> & RouterLinkProps
+  ) => <ListItemButton component={RouterLink} {...props} />;
+
+  const [filter, setFilter] = useState<string>('');
+  const filterVCs = (virtualContributor: VirtualContributorProps) => {
+    const lowerCaseFilter = filter.toLowerCase();
+
+    return (
+      virtualContributor.profile.displayName.toLowerCase().includes(lowerCaseFilter) ||
+      virtualContributor.profile.tagline.toLowerCase().includes(lowerCaseFilter)
```
bobbykolev, line 47: Why are you searching in the tagline? I think it should be only on `displayName` (unless specified otherwise).

src/domain/community/contributor/CommunityContributorsBlockWide/CommunityVirtualContributorsBlockWide.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { ContributorType } from './CommunityContributorsBlockWideContent';
+import Loading from '../../../../core/ui/loading/Loading';
+import usePlatformOrigin from '../../../platform/routes/usePlatformOrigin';
+
+interface CommunityContributorsBlockWideProps {
+  virtualContributors: VirtualContributorProps[];
+  isLoading?: boolean;
+}
+
+const COMPACT_VIEW_ITEMS_LIMIT = 3 * 8;
```
bobbykolev, line 21: Please leave a comment about why 3 * 8 - e.g. // 3 rows on Desktop

src/domain/community/contributor/CommunityContributorsBlockWide/CommunityVirtualContributorsBlockWide.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const isSmallScreen = useMediaQuery<Theme>(theme => theme.breakpoints.down('lg'));
+  const columns = useColumns();
+
+  const filterFn = (filter: string[]) => (element: VirtualContributorProps) => {
+    return (
+      filter.length === 0 ||
+      filter.some(
+        term =>
+          element.profile.displayName.toLowerCase().includes(term.toLowerCase()) ||
+          element.profile.tagline.toLowerCase().includes(term.toLowerCase())
```
bobbykolev, line 37: Again the filtering by tagline seems redundant. 

Also, the `filterFn` could be better named because the following is not pretty:
.filter(filterFn(filter))

It would look a bit better if it's `virtualContributors.filter(matchesNameFilter(searchTerm))` or something like that.

src/domain/journey/common/AboutPageContainer/AboutPageContainer.tsx
```diff
@@ -85,6 +89,12 @@ const AboutPageContainer: FC<AboutPageContainerProps> = ({ journeyId, children }
   });
 
   const memberProfile = membersData?.lookup.space?.profile;
+  const virtualContributors = membersData?.lookup.space?.community.virtualContributors.filter(
+    vc => vc.searchVisibility === SearchVisibility.Public
```
bobbykolev, line 93: We've discussed that filtering briefly. I think we should include the visibility per Account/Private on that level too.
Feel free to leave it like that but please mark it in the ticket.

src/domain/journey/space/SpaceCommunityPage/SpaceCommunityPage.tsx
```diff
@@ -89,6 +96,11 @@ const SpaceCommunityPage = () => {
 
   const [isActivitiesDialogOpen, setIsActivitiesDialogOpen] = useState(false);
 
+  const virtualContributors: VirtualContributorProps[] =
+    data?.space.community.virtualContributors?.filter(vc => vc?.searchVisibility === SearchVisibility.Public) ?? [];
+
+  const hasReadPrivilege = data?.space.authorization?.myPrivileges?.includes(AuthorizationPrivilege.Read);
```
bobbykolev, line 102: I'm not sure whether the effort is worthy performance-wise but in case hasReadPrivilege === false, then there's no need of the assignment and filtering on line 100 (on other occasions as well).
It could be optimized like:

```
let virtualContributors: VirtualContributorProps[] | undefined;

const hasReadPrivilege = data?.space.authorization?.myPrivileges?.includes(AuthorizationPrivilege.Read);

if (hasReadPrivilege) {
 virtualContributors = data?.space.community.virtualContributors?.filter(vc => vc?.searchVisibility === SearchVisibility.Public);
}

**Comment by bobbykolev** [comment]
I think for these blocks with VCs in the community, the list shouldn't be filtered only by 'Public'. That way the VC on that particular account are not displayed.
image
image

@techsmyth , please confirm the filtering - 1. only public; 2. only non-hidden; 3. no filtering;

File: src/domain/journey/common/AboutPageContainer/AboutPageContainer.tsx (hunk 8 above)
Comment (EmiliaPavlova) [reply src/domain/journey/common/AboutPageContainer/AboutPageContainer.tsx:93]: There is an explicit bullet to not show private vcs here. It's not clear what you want here.

File: src/domain/community/contributor/CommunityContributorsBlockWide/CommunityVirtualContributorsBlockWide.tsx [hunk 10]
```diff
@@ -73,7 +66,7 @@ const CommunityVirtualContributorsBlockWide = ({
                       avatar={vc.profile.avatar?.uri ?? ''}
                       displayName={vc.profile.displayName}
                       tooltip={{ tags: [] }}
-                      url={vc.profile.url.substring(origin.length, vc.profile.url.length) ?? ''}
+                      url={buildVirtualContributorUrl(vc.nameID)}
```
bobbykolev, line 69: Please, request the profile { url } instead of nameID. No need of `buildVirtualContributorUrl` just pass the vc.profile.url.

src/domain/journey/common/AboutPageContainer/AboutPageContainer.tsx
bobbykolev: I'll update the ticket, thanks!

bobbykolev:
> @techsmyth , please confirm the filtering - 1. only public; 2. only non-hidden; 3. no filtering;

As agreed we're showing all non-hidden member VCs.

techsmyth:
@bobbykolev agreed re display.

bobbykolev (approved):

Issue #6545, opened 2024-07-08
BUG: client redirecting to dashboard on user route


techsmyth:
**Describe the bug**
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNXErQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5ee3acfa6cf49a9a7e65ca90ab4f4d0c665eaae6/image.png" alt="image.png" />

<img width="1781" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2krQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f5f94e40d521a3065f12c05732995e991c172bbb/image.png" alt="image.png" />

**To Reproduce**

1. Navigate to [domain]/user e.g. https://acc-alkem.io/user

**Expected behavior**
The top-level routes to be protected on the client. 
1. do not allow creation of a space with nameID of one of the Platform Top Level Routes;
2. in case of access to one of them like '/user' -> navigate to home or 404, not to '/user/dashboard' with broken space layout.

techsmyth:
@me-andre for now please enforce it only the client, I will add a separate issue to reserve nameIDs on the space level

Comoque1:
Tested OK on DEV env.

PR #6556, opened 2024-07-09
Introduced restricted top level routes

me-andre:
https://github.com/alkem-io/client-web/issues/6545

- [x] proper Route handling of TopLevel routes like '/user' loading 404 instead of empty space;
- [x] check on space creation for validity of the nameID - couldn't be one from the top level routes;


src/main/routing/reservedTopLevelRoutePaths.ts
```diff
@@ -0,0 +1,3 @@
+const reservedTopLevelRoutePaths = ['admin', 'identity', 'user', 'vc', 'organization'];
```
bobbykolev, line 1: I think we should add all existing static paths like - contributors, about, profile, forum etc.

techsmyth: Agreed, just makes it safe now.

src/main/routing/TopLevelRoutes.tsx
```diff
@@ -62,10 +60,10 @@ export const TopLevelRoutes: FC = () => {
...
         <Route
-          path={ROUTE_HOME}
+          path={TopLevelRoutePath.Home}
           element={
             <NonIdentity>
-              <WithApmTransaction path={ROUTE_HOME}>
+              <WithApmTransaction path={TopLevelRoutePath.Home}>
                 <HomePage />
               </WithApmTransaction>
             </NonIdentity>
```
bobbykolev, line 69: I'd suggest the Space route below, to be placed on the bottom of the list after all static/reserved paths.

bobbykolev (approved):

PR #6585, opened 2024-07-12
Lazy loading Excalidraw

me-andre:

Reduces prod bundle from 8.2Mb to 5.8Mb without sourcemaps and from 21 to 7 with sourcemaps.

### Additional context

When a Whiteboard is opened for the first time, the user observes a spinner for a second or two. Subsequent openings of this or other Whiteboards don't have this delay.


valentinyanakiev (approved):
Works as expected - tested locally.

PR #6614, opened 2024-07-17
Fix server 4251: Read My created spaces from its own field

ccanos:
Goes together with server PR: https://github.com/alkem-io/server/pull/4273

I've just removed the param `showOnlyMyCreatedSpaces` and have added a field to the `me` query that returns directly my created spaces instead of passing through activities.


bobbykolev (approved):
Looking good @ccanos 👏 
Please, make sure it's not merged before the server changes get in.

Issue #6632, opened 2024-07-18
BUG: Space>Community page: The community block is not filled


SimoneZaza:
**Describe the bug**
The community block is not filled with available users. Therefore all large communities seem to be much smaller then they are. See as an example the vu (117 users) below. Decreases the power of showing 'this very big community' in demo's ;)
 
<img width="1041" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnJ2QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a5889862f62cbcaa479e8561bdbdd24f5314cb35/image.png" alt="image.png" />
Same community, when I click on "Show all"
<img width="1546" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnZ2QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--06137fd586b349714f0c9da77b11affe72f6b95a/image.png" alt="image.png" />

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

**Areas that will be affected**
_Optional, to be added by developer if relevant_

Comoque1:
Tested OK on ACC env.

PR #6664, opened 2024-07-24
Fix a grid issue with contributors cards

bobbykolev:
https://github.com/alkem-io/client-web/issues/6632

After we introduced the 3 by 9 division of the main grid (instead of 4 + 8), the layout of the contributors had been broken. 
Instead of showing 3 rows of 8 contributors, it started rendering 9 in a row (useColumns() hook). 
We ended up showing 3 * 8 contributors in a grid of 3 * 9. 
24 is dividable by 2 and 4, and has a better representation on every resolution. So the fix sets the columns on Desktop to 8, regardless of the useCollumns hook. The logic remains the same for smaller res.


EmiliaPavlova (approved):

bobbykolev (commented):
Thanks for the improved implementation @me-andre 🙏
Let's merge!

Issue #6669, opened 2024-07-25
Route information with Sentry Errors 


bobbykolev:
## Description

Let's enhance our error data with router information, specifically the last URL path before the actual error. 

## Acceptance criteria

- [ ] The Sentry error should contain router information (TBD)
- [ ] …

## Additional Context

Screen shots / UX designs / data that are relevant.

## Areas that will be affected 
 
_To be added during the refinement_

ccanos:
@Comoque1 when I started to work on this, it seemed to me that dev environment is not logging to Sentry, let me know if you need help to enable logging or keep that in mind and test this in sandbox or acceptance...
A couple of possible tests:
- create a post with a link in the description, a link to something that doesn't exist, like a wrong space name or an entirely wrong url in the platform. You can start navigating, have in mind to go from space A to B, then open the callout, then click on the post, the click on that wrong link: then you come here:
<img width="1445" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3RFREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5cbe56d12f3e101cd53ce03771ab2b85777c9396/image.png" alt="image.png" />
and see the last 5 links you passed by in the history:
<img width="1216" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3hFREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9f2089bdc6db7eda9e68cc6cea4abb776add2a2b/image.png" alt="image.png" />

- another possible test it to create a post in a callout, open the callout, delete the post from another tab and then open the post
Things like that, that trigger the error page

Issue #6566, opened 2024-07-11
All space routes that fail to resolve to be logged in sentry


techsmyth:
## Description

As a product owner I want to know who has fixed old links to subspaces via URL paths that need updating

## Acceptance criteria

- [ ] Log all URLs for spaces that fail to resolve in sentry, with details of the originating URL

## Additional Context

Screen shots / UX designs / data that are relevant.

## Areas that will be affected 
 
_To be added during the refinement_

PR #6670, opened 2024-07-25
Log Sentry 404 in the console and add tag to it

bobbykolev:
https://github.com/alkem-io/client-web/issues/6669
We're logging Sentry 404 already. We're trying to send the current URL and the referrer. However, the referrer is not working as we would want - sending the SPA path.

In this PR the error is enhanced with a tag and we now log it in the dev console so we can monitor what errors are captured by Sentry during development. 

We'll need special implementation that captures every url/path in the sessions so in case of an error, we can retrieve the last visited path.


valentinyanakiev (approved):

valentinyanakiev (approved):

bobbykolev:
Part of https://github.com/alkem-io/client-web/issues/6566

PR #6671, opened 2024-07-25
 My Memberships dialog: UX update

EmiliaPavlova:
Redesign of My Membership:
- Expandable list view for dialog
- Per list item: avatar, title, tagline, and the word(s) admin, lead when relevant
- Sort spaces on activity, subspaces on displayname (as is currently the case in this dialog)
- Show everything collapsed by default
- Use spinner for loading indication
- Mobile: see design, show admin/lead label underneath title/tagline instead of next to it
- Mobile if easy: truncate title at 2 lines, tagline at 2 lines
- Keep current footer message

Design: https://www.figma.com/file/noZ6QjMORE2nNkhccSc9pG/Homepage?type=design&node-id=1491%3A9141&mode=design&t=OKMPurcBo29buoSW-1


src/main/topLevelPages/myDashboard/myMemberships/MyMembershipsSpaceCard.tsx
```diff
@@ -0,0 +1,103 @@
...
+              endIcon={isExpanded ? <KeyboardArrowUpIcon /> : <KeyboardArrowDownIcon />}
+              sx={{ visibility: subspaces?.length ? 'visible' : 'hidden' }}
+              area-label={isExpanded ? t('buttons.collapse') : t('buttons.expand')}
+            />
+          </Gutters>
+        </Gutters>
+      </GridItem>
+      {isExpanded &&
+        subspaces?.map(subspace => (
+          <MyMembershipsSpaceCard
```
bobbykolev, line 88: This nesting looks dangerous, but it works well for our structure. 

I suggest rethinking the name of the component as it's not just a Space card but as it renders its children subspaces, it could be something like MyMembershipsSpaceTree.

bobbykolev:
This dialog looks far better now. Good job 👏

Please, make sure we're displaying the custom avatar on a space level (on subspace levels it's fine). 
image

bobbykolev (dismissed):

src/main/topLevelPages/myDashboard/myMemberships/MyMembershipsDialog.tsx
```diff
@@ -66,57 +58,14 @@ const MyMembershipsDialog = ({ open, onClose }: MyJourneysDialogProps) => {
...
-                tagline={space.profile.tagline}
-                journeyUri={space.profile.url}
-                tags={space.profile.tagset?.tags ?? []}
-                membersCount={getMetricCount(space.metrics, MetricType.Member)}
-                spaceVisibility={space.account.license.visibility}
-              />
-              <GridItem columns={9}>
-                <Gutters row disablePadding flexGrow={1} flexWrap="wrap">
-                  <GridProvider columns={8}>
-                    {space.subspaces?.filter(isJourneyMember).map(subspace => (
```
bobbykolev, line 110: ⚠️ This check is important as if you're part of a Space, it doesn't mean you are part of the subspaces. In this dialog we should show only the spaces/subspaces/subsbuspaces the user is actually a member of.

In case of a private subspace (that the user is not a member of) currently the entire query is failing.

bobbykolev: fixed with 81d71b5152ec37215b7616d6ae26e9bf9921eb88

bobbykolev (approved):

Issue #6621, opened 2024-07-18
VC Create flow Step 3: Try out dialog


SimoneZaza:
## Description

As a user, I want the create flow to align with my understanding and wishes for my VC. I want to learn about the options through the create flow, without getting too much information. The way it works should teach me. 

## Acceptance criteria

- [ ] This dialog has an example post in which they interact with the VC
- [ ] @<VC name> is prefilled in the input field
- [ ] In the background, the page changed to the Personal Space
- [ ] When the user closes the dialog, they remain on the Personal Space page
- [ ] The example callout is also on the space page, on the dashboard

## Additional Context

User flow: https://alkem.io/building-alkemio-org/challenges/improvingtheusere-6400/collaboration/vccreateflowwith3typ

UI Designs: https://www.figma.com/design/gV8GGWh99QRKfM65lwedrE/Dashboard-Update?node-id=482-207&t=jCXmCypVZY8hWnEi-1
<img width="1045" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMVh1QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8f8f3c6d68ffb4cb2ff72258c23afcb880caadc5/image.png" alt="image.png" />

## Areas that will be affected 
 
_To be added during the refinement_

bobbykolev:
Tech suggestion:
Use route param or localStorage (better) to show this dialog on space landing. In the previous step of the flow save the value in LS.

PR #6681, opened 2024-07-30
Try your VC dialog 

bobbykolev:
https://github.com/alkem-io/client-web/issues/6621

- [x] Dialog showing on a space landing after the VC creation flow is completed;
- [x] Post comment callout created in the board;
- [x] Post details shown (translated post title & description);
- [x] Visibility dependent on LocalStorage (could be replaced if needed);
  - you can use  the following to set the VC name `addVCCreationCache(vcName);`;
  - or `localStorage.setItem('TRYVC', 'Virtual Assistant 101');` for testing purposes (in the browser's dev console);
- [x] Dynamic population of the VC name in the text;

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -2970,5 +2970,15 @@
       "updateProfile": "Update profile",
       "addKnowledge": "Add knowledge"
     }
+  },
+  "createVirtualContributorV2": {
+    "trySection": {
+      "title": "🎉 Try it out!",
+      "subTitle": "Your Virtual Contributor is successfully created and we added it to your personal Space <tooltip><icon /></tooltip>. You can now interact with it by typing @<name VC> in a comment on any post in your Space. Give it a try in the preview below!",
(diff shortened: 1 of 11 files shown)


bobbykolev:
@ccanos please revise the logic, especially on the:
- For callout creation, I had to set Timeout of 5s to ensure the creation was completed every time. For some reason, the createCalloutHook.handleCreateCallout was called but there was no creation query afterwards.

ccanos (approved):
@bobbykolev Please, take a look at my commits, I got to solve it.
The problem was that `useCalloutCreation` cannot work immediately, it has to retrieve the permissions and the `collaborationId` with `useCollaborationAuthorization`, so you cannot use it until `loading` is false. 
I didn't use `loading` itself because it also changes when the hook is creating a callout, so I was getting the effect executed infinite times (on every `loading` change).
I have added a new flag `canCreateCallout`, that is a duplicate of the permission, but I think it's a good to have.
I have also changed the variable name postCallout to demoCallout because I thought it was more clear but feel free to change it back or set another name

bobbykolev:
> @ccanos thank you for the suggestions and fixes. I agree with all of them 🙏

Issue #6623, opened 2024-07-18
Account tab (2) in users' Settings


SimoneZaza:
## Description

As a user, I want to be able to manage my account and the entities it contains

## Acceptance criteria

- [ ] Same as the Organization account tab - https://github.com/alkem-io/client-web/issues/6440:
- Spaces
- Virtual Contributors
- + Innovation packs and custom homepages as they are moved under an account and a user can be a host;
- [ ] Reuse Create a Space existing functionality;
- [ ] Create Innovation Pack functionality (extracted and implemented from https://github.com/alkem-io/client-web/issues/6624)

UPDATE: 5.8.24
As InnoPacks and Hubs are part of the account, they will be visible on the user's profile if any.
As the InnoPack section has a Create button, it's shown even if you don't have innovation packs yet. 

- [ ] Remove the create InnoPack option in the Global Administration. 

## Additional Context

Figma (disregard the designs that are not part of this issue): https://www.figma.com/design/l0hLyZCBs5cxhpp5vUPAte/Account-page?node-id=186-106&t=AhyNIs3ukaS3wWt7-1

## Areas that will be affected 
 
_To be added during the refinement_

bobbykolev:
What was altered and could be extracted for refinement or bugs:
- There's no support on the list of items for the space name (on the right for all entities but the spaces);
- The InnoPack create button is visible (the entire section) even if you 1. don't have other packs and 2. not only for organizations (as requested);

Comoque1:
Tested OK on DEV env.

PR #6698, opened 2024-08-02
My Account Tab

ccanos:
As a user I want to be able to manage my account and the entities it contains

https://github.com/alkem-io/client-web/issues/6623


src/core/i18n/en/translation.en.json
```diff
@@ -1836,7 +1836,11 @@
             }
           },
           "account": {
-            "description": "",
+            "description": "Here you can see details about things hosted by your account.",
+            "hostedSpaces": "Hosted Spaces",
+            "virtualContributors": "Virtual Contributors",
+            "innovationPacks": "$t(common.innovation-packs)",
+            "customHomepages": "My Custom Homepages",
```
bobbykolev, line 1843: Design says "Your" ... Custom Homepages

src/core/i18n/en/translation.en.json
```diff
@@ -1974,6 +1978,7 @@
         }
       },
       "innovation-packs": {
+        "create": "Create InnovationPack",
```
bobbykolev, line 1981: A space between "Innovation"  and "Pack", I guess.

src/domain/collaboration/InnovationPack/InnovationPackCardHorizontal/InnovationPackCardHorizontal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    url: string;
+  };
+  calloutTemplatesCount?: number;
+  communityGuidelinesTemplatesCount?: number;
+  innovationFlowTemplatesCount?: number;
+  postTemplatesCount?: number;
+  whiteboardTemplatesCount?: number;
+}
+
+export const InnovationPackCardHorizontalSkeleton = () => (
```
bobbykolev, line 30: InnovationPackCardHorizontalSkeleton && InnovationHubCardHorizontalSkeleton looks the same, can we use just one of them for both skeletons?

src/domain/journey/space/createSpace/CreateSpaceDialog.tsx
```diff
@@ -41,8 +40,7 @@ interface FormValues extends SpaceEditFormValuesType {
 const CreateSpaceDialog = () => {
   const handleClose = useBackToStaticPath(ROUTE_HOME);
```
bobbykolev, line 41: If you close the dialog on user->account, it lands you back on the dashboard. Ideally, it should either accept a path or just close itself without redirection.

bobbykolev (commented):
A couple of minor things, see the comments @ccanos, nothing critical there.

However, as discussed:
- the user query is failing locally (it will fail on anther envs as well) if the user has an account without associated spaceId. I don't know how this is happening but it exists and it's blocking the access to the page;
- We'll extract a ticket for consolidation of the create space dialogs;

src/core/i18n/en/translation.en.json
ccanos: 😅 good catch

src/domain/journey/space/createSpace/CreateSpaceDialog.tsx
ccanos: I'd like to keep them separated in case their visualizations divert more, and together with their non-skeleton card... But I have done now the InnovationPack skeleton a bit different to have the excuse :D

src/domain/collaboration/InnovationPack/InnovationPackCardHorizontal/InnovationPackCardHorizontal.tsx
ccanos: I'd like to keep them separated in case their visualizations divert more, and together with their non-skeleton card... But I have done now the InnovationPack skeleton a bit different to have the excuse :D

src/domain/journey/space/createSpace/CreateSpaceDialog.tsx
ccanos: This was a bad copy&paste, good catch

src/core/i18n/en/translation.en.json
ccanos: It's because Rene's message :)

ccanos:
Indeed @bobbykolev we'll need to join those dialogs, I have created #6700  to not forget about it. We'll take a look to that query next week. thanks for the review.

bobbykolev:
- We should add a check for the space creation option: `user.hasPlatformPrivilege(AuthorizationPrivilege.CreateSpace)`. Currently, it's available on every UserAccount tab and ends in an infinite loading screen with an error if the user lacks this privilege.

bobbykolev (approved):

PR #6701, opened 2024-08-05
Client updated to work with hierarchical space memberships query

techsmyth:
Dependant on: https://github.com/alkem-io/server/pull/4354 
Merging needs synchronization. ⚠️

Codegen is compiling
Code is compiling

To test the functionality / results...

To check:
- please look at the todos to see if the client depended on a flat list previously. 
- filtering on spaces explorer page is not working


bobbykolev:
- [x] MyMemberships logic was fixed, no it works far faster and better 👏
- [ ] Latest contributions need a fix; now they can be filtered only by top-level spaces, but all levels should be supported.

valentinyanakiev (approved):

PR #6705, opened 2024-08-05
Account tab fixes

bobbykolev:
https://github.com/alkem-io/client-web/issues/6623

- [x] Custom space avatars not showing;
- [x] Innovation Pack creation - redundant fields - organization, visibility;

Note that the reusable JourneyCardHorizontal was changed. In my opinion better OR check added to support an empty string besides null and undefined (causing the first issue).

On the second point, there was one more request from Evgeni - to remove the Organization Provider on the Innovation Pack settings page, as well. I'll leave that for a separate ticket.


ccanos (approved):

ccanos (changes requested):
fixing something before merging

ccanos (approved):

Issue #6624, opened 2024-07-18
Account tab 3: create from page


SimoneZaza:
## Description

As a user, I want to be able to create new things from my account page as well

## Acceptance criteria

- [ ] Plus button
- Bottom left
- Only visible for users with the create privilege for the corresponding entity
---
- [ ] Spaces:
- Open create flow as it currently is there from the dashboard: implemented with https://github.com/alkem-io/client-web/issues/6623
- Privilege required: CREATE_SPACE on Account entity
- limit to 3 (if not a GA - inconsistent with the home's dashboard @SimoneZaza);
- [ ] Virtual Contributors:
- Open create flow as it currently is there from the dashboard
- Each user / org can only have one account
- Privilege required: CREATE_VIRTUAL_CONTRIBUTOR on Account entity
- limit to 3 (if not a GA, as in the home's dashboard);
- [ ] Innovation Packs:
- Use current create fields (from global administration, see screenshot below), but move it to a dialog: implemented with https://github.com/alkem-io/client-web/issues/6623
- Rename SAVE into CREATE
- Privilege required: CREATE_INNOVATION_PACK on Account entity
- Create button opens innovation pack profile settings page (as it currently does)
- [ ] Custom Homepages:
- Privilege required: CREATE_INNOVATION_HUB on Account entity
- Create button opens innovation pack profile settings page (as it currently does)

## Additional Context
Higher estimate because of the many flows that need to work

Figma (disregard the designs that are not part of this issue): https://www.figma.com/design/l0hLyZCBs5cxhpp5vUPAte/Account-page?node-id=186-106&t=AhyNIs3ukaS3wWt7-1

Current create fields Innovation pack:
<img width="1089" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNC91QlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--59332cb26ea4ab11e888f51f04743dcd8d1e768a/image.png" alt="image.png" />

## Areas that will be affected 
 
_To be added during the refinement_

techsmyth:
Moved back to refined column as description changed slightly

bobbykolev:
Note that **Create Space** and **Create Innovation Pack** are available on the Users' Account Tabs (implemented as part of https://github.com/alkem-io/client-web/issues/6623).

What remains:
- Creation in the Orgs' Account Tabs;
[...]

PR #6723, opened 2024-08-07
Organization AccountTab

ccanos:
Goes together with  https://github.com/alkem-io/server/pull/4370
For now it is practically the same as UserAccountPage organized in a different way


src/domain/community/user/pages/UserAccountPage.tsx
```diff
@@ -47,7 +48,7 @@ export const UserAccountPage: FC<UserAccountPageProps> = () => {
 
   const { spaceIds, virtualContributors, innovationPacks, innovationHubs } = useMemo(
```
bobbykolev, line 49: Not related directly to the review but maybe declare `const accounts = data?.user?.accounts;` and then use it in the memo instead of repeating `data?.user?.accounts.`

src/core/i18n/en/translation.en.json
```diff
@@ -1933,6 +1933,7 @@
         },
         "notifications": {
           "space-created": "Space created successfuly.",
+          "errorCreatingSpace": "Error creating Space.",
```
bobbykolev, line 1936: It could go like that but the wording could be improved:
"An error occurred while creating the Space."

bobbykolev (dismissed):
Looking good @ccanos , nothing critical in the comments to be fixed.
We lack creation options but those could be carefully done with the separate the following task:
https://github.com/alkem-io/client-web/issues/6624

bobbykolev (approved):

PR #6730, opened 2024-08-08
VC flow - fix space name collision

bobbykolev:
Fixing a bug when creating space for a second time after cancelation.
 
STR:
1. cancel the flow after space creation (on the callouts screen);
2. then start the creation again;
3. A space with the same name and nameID is created;
4. On navigation to the space (after saving the post callouts), there are errors and the space doesn't load correctly.

With the change in the PR, we're increasing the index on deleting a space. That way the next Space has a different nameID, and there are no collisions. 

This fix is nice-to-have as we notify the user that the page needs to be reloaded on cancelation.

### DIFF
--- a/src/main/topLevelPages/myDashboard/newVirtualContributorWizard/useNewVirtualContributorWizard.tsx
+++ b/src/main/topLevelPages/myDashboard/newVirtualContributorWizard/useNewVirtualContributorWizard.tsx
@@ -56,11 +56,11 @@ interface NewVirtualContributorWizardProps {}
 
 // generate name for the space/subspace based on the VC name
 // index is needed in case of canceling the flow. Creation with the same name/nameID leads to issues accessing it later
-const generateEntitiName = (name: string, entityPostfix: string, index: number = 0) => {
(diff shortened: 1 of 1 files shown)


src/main/topLevelPages/myDashboard/newVirtualContributorWizard/useNewVirtualContributorWizard.tsx
```diff
@@ -50,8 +54,13 @@ interface useNewVirtualContributorWizardProvided {
 
 interface NewVirtualContributorWizardProps {}
 
-const generateSpaceName = (name: string) => `${name}'s Space`;
-const generateSubSpaceName = (name: string) => `${name}'s BoK`;
+// generate name for the space/subspace based on the VC name
+// index is needed in case of canceling the flow. Creation with the same name/nameID leads to issues accessing it later
+const generateEntitiName = (name: string, entityPostfix: string, index: number = 0) => {
```
EmiliaPavlova, line 59: Typo here

EmiliaPavlova (dismissed):

PR #6742, opened 2024-08-09
Fix VC name validation

EmiliaPavlova:


hero101 (changes requested):
yup is only concerned with validation.
the actual data given as an input to the server, can still be corrupted.

I was still able to create the following VC through the client:
image

The nameID generation must also handle inputs with too much spaces inside.

hero101 (approved):

PR #6755, opened 2024-08-12
Organization Account tab fixes

ccanos:
- Fixed Clicking on VC profile (refactor ContributorCardHorizontal and all its usages, now url is in the profile).
- Hide innnovationPacks if there are none
- Fix missing icons from the InnovationPacks cards
- Only link to the InnovationHub settings if you're an admin

### DIFF
--- a/src/domain/innovationHub/InnovationHubCardHorizontal/InnovationHubCardHorizontal.tsx
+++ b/src/domain/innovationHub/InnovationHubCardHorizontal/InnovationHubCardHorizontal.tsx
@@ -4,9 +4,10 @@ import Avatar from '../../../core/ui/avatar/Avatar';
 import RouterLink from '../../../core/ui/link/RouterLink';
 import { Caption, CardTitle } from '../../../core/ui/typography';
 import OneLineMarkdown from '../../../core/ui/markdown/OneLineMarkdown';
-import { SpaceVisibility } from '../../../core/apollo/generated/graphql-schema';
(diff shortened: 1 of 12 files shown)


src/domain/innovationHub/InnovationHubCardHorizontal/InnovationHubCardHorizontal.tsx
```diff
@@ -62,8 +63,13 @@ const InnovationHubCardHorizontal = ({
   profile: { displayName, description, url, banner },
   ...spaces
 }: InnovationHubCardHorizontalProps) => {
+  const { user: { hasPlatformPrivilege } = {} } = useUserContext();
+  const isPlatformAdmin = hasPlatformPrivilege?.(AuthorizationPrivilege.PlatformAdmin);
+
+  const SelectComponent = () => (url && isPlatformAdmin ? { component: RouterLink, to: url } : { component: Box });
```
bobbykolev, line 69: this is more of a `getComponentProps` function. It could be a component if you pass it to the component prop instead of spread.

bobbykolev (dismissed):
Looking good, @ccanos

src/domain/innovationHub/InnovationHubCardHorizontal/InnovationHubCardHorizontal.tsx
ccanos: totally right, I forgot to clean this up, it doesn't even need to be a function or to return anything if it's not a link it's just a Box already

bobbykolev (approved):

Issue #6762, opened 2024-08-14
BUG: When BETA user create a Space with Plus plan, he has it displayed as free


Comoque1:
**Describe the bug**

**To Reproduce**

1. Go to https://acc-alkem.io/user/august14--14-2024/settings/account (user profile settings / account  of a Beta tested)
2. Create 2 plus spaces
3. Navigate to the spaces account - they are displayed as free
<img width="1335" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBek5YQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d5af25ac63d56ed7165b583123f7f1c223902901/image.png" alt="image.png" />

**Expected behavior**
Spaces with Plus license are displayed as such

Comoque1:
Tested OK on ACC env.

PR #6763, opened 2024-08-14
Temporary hide the space creation button under Account tab

bobbykolev:
https://github.com/alkem-io/client-web/issues/6762
Because of тхе issue where users can select different License Plans but their choice is ignored and all are considered Free.


ccanos (approved):

bobbykolev:
We're not hiding the option for now.

PR #6767, opened 2024-08-14
Build InnovationHub url from subdomain and not point to the IH settings

ccanos:


bobbykolev (approved):
Looking good, tested locally, now the InnoHubs are loading correctly.

PR #6770, opened 2024-08-14
Keep selected collapsed/expanded state of the subspaces menu

EmiliaPavlova:
Keeping the selected state for all subspaces.
Renaming the prop isExpanded - having isExpanded(true) for collapsed and isExpanded(false) for expanded is very confusing.


bobbykolev (approved):
Changes are nice and clean, the implementation works well. 


I'd suggest extracting the LS key in a const:

`const MENU_STATE_KEY = 'menuState';`

Also, by convention the keys of the enum should be uppercase:
```
enum MenuState {
  EXPANDED = 'expanded',
  COLLAPSED = 'collapsed',
}
```

But those suggestions are nice to have. Feel free to merge as is.

bobbykolev (approved):

Issue #6793, opened 2024-08-21
BUG: VC flow finishes with an error


ccanos:
**Describe the bug**

**To Reproduce**

1. Follow the vc creation flow (add knowledge or existing subspace)
2. After VC creation an error is thrown and the VC is not added to the BoK community.

**Expected behavior**
No errors, successful flow.

**Screenshots**

**Additional context**
See details in the linked PR

**Areas that will be affected**
_Optional, to be added by developer if relevant_

Comoque1:
The issue fails with another (authorization) error now.
Moving it back to backlog.

bobbykolev:
Another error extracted here:
https://github.com/alkem-io/server/issues/4445

PR #6798, opened 2024-08-21
VC creation flow fixes

bobbykolev:
- [x] Fixing https://github.com/alkem-io/client-web/issues/6793 by storing the parentCommunityId and adding the VC to it;
- [x] Translations - fix of the VC creation success msg; auto-generated vision & tagline for subspace are now in the translations;
- [x] `spaceId` & `mySpaceId` were ambiguous making the code hard to follow. Fixed.
- [x] ⚠️ UX change - on Existing knowledge and subspace creation, navigate the user to the Space (not the subspace as it was);
- cleanup;
 

Note, that it's not ideal but currently we're adding the VCs to the first space in the account (creating subspaces there):
`selectedExistingSpaceId: mySpaces?.[0]?.id,`
I don't know how we can improve that, atm.

### DIFF
--- a/src/core/apollo/generated/apollo-hooks.ts
+++ b/src/core/apollo/generated/apollo-hooks.ts
@@ -16722,6 +16722,66 @@ export function refetchSpaceUrlQuery(variables: SchemaTypes.SpaceUrlQueryVariabl
   return { query: SpaceUrlDocument, variables: variables };
 }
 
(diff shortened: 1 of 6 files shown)


ccanos (changes requested):
Just a small (I hope) change if you can, 
if not we can write it down for later, looks to be working fine

ccanos (approved):

PR #6831, opened 2024-08-28
Templates refactor

ccanos:
# TODO - Create tasks for these
 - Mappings file can be rewritten using plain Typescript, I think the use of immer's produce has been a mistake,  but for now the file is clear and works... it's just the types are not 100% consistent and Typescript loses track.
 - Neil has created a CreateInput service, a field in the GraphQL API, like lookups, that is returning a CreateCalloutInput, CreateCommunityGuidelinesInput,... Objects that can directly be used to call the mutations to create a Callout or to create a CalloutTemplate.
  That may (*it will*) simplify the mappings file.
  I initally implemented the Template Import from IP to Space copying the data and mapping it, it's in TemplatesAdmin, and it's using ImportTemplateInSpace.graphql. Please try to implement it using useCreateInputFromTemplate, a separate hook.
 - Also a new type of template is added, CollaborationTemplate. finish the implementation.
 - Space templates will be implemented soon, there should be a task already for that.

# What to test:

- I would open two browsers, one in acc and one in develop and go through everywhere where there's a template and comparing.
- Styles have changed, there were lots of different components doing the same and now they are unified so many styles have changed, check that everything is relatively nice, or at least consistent. Maybe I have broken some responsiveness/mobile view, but I think I actually have fixed them using a single component that handles responsiveness correctly.

- Images, card visuals for the templates, I haven't been able to test that properly locally.
  - whiteboard templates get the whiteboard screenshot as banner normally,
- Also things uploaded to references, you may find some missing clips (where you can't attach a file), that means I have removed the StorageContext

- Places where there are templates, or template related things:
  - Global Admin => Spaces settings => Templates
    There you can View, Edit, Create, Delete and Import from platform library
  - Global Admin => Innovation Packs => Innovation pack
  - Space => Create callout => Whiteboard => Edit => Find template
    In general in during any whiteboard you're editing you should be able to import a template
    Templates from the space should be available first, then templates from the platform
  - Space => Create callout => Call for posts => Find template
  - Space => Create callout => Call for Wbs => Find template
  - Space => Create callout => Have a look in the library!
  - (Same for subspaces, please check subspaces and sub-subspaces to check if they are loading correctly templates from their parent space)
  - Subspace => InnovationFlow settings
  - Space => Settings => Community => CommunityGuidelines


src/core/i18n/en/translation.en.json
```diff
@@ -512,16 +512,32 @@
...
+        "InnovationFlow": "Innovation Flow",
+        "InnovationFlow_plural": "Innovation Flow Templates",
+        "Callout": "Collaboration Tool Template",
+        "Callout_plural": "Collaboration Tool Templates",
+        "CommunityGuidelines": "Community Guidelines Template",
+        "CommunityGuidelines_plural": "Community Guidelines Templates",
+        "Collaboration": "Collaboration Template",
+        "Collaboration_plural": "Collaboration Template"
+      },
+      "templateType": {
```
bobbykolev, line 528: Do we need both `templateTypes` and `templateType` (different keys, same values)?

bobbykolev:
After the initial pass, things are looking pretty good 👏

I'll list some of the issues that I've spotted (some of them should be fixed in https://github.com/alkem-io/client-web/pull/6855):
- [x] **Creation of WB template was broken** due to passing `profile` and the API expecting `profileData`;
- [x] ⚠️ **Space setting -> community** page was crashing (probably related to the fix of the previous point) when there's no 'profile';
- [ ] **Space settings -> community** page - CGL templates cannot be used. Click on USE - nothing happens.
- [ ] **Visuals are missing** in some of the templates. Interestingly, the server is giving empty URIs, so it's probably related to the preview generation;
- [ ] ⚠️**sub/subsub space creation is broken** - no default InnovationHub;
- [ ] **Library** - filter chips are broken (on/off for all options);
- [ ] ↘️ **Space Settings -> Templates** - you can add the same template multiple times;

bobbykolev (approved):

Issue #6808, opened 2024-08-23
Deleting resources from User/org account tab


SimoneZaza:
## Description

As a user, I want to be able to delete my VCs, Innovation packs, and custom homepages
As a user, I want to be able to delete my Spaces from my account tab

### Space
- [ ] Add "more" icon button on Space card
- [ ] This button opens a little menu with -for now only- a delete option
- [ ] Clicking delete opens the same confirmation dialog as the one on the Space>Settings>Account>Delete:
- Title: Confirm [Space/Virtual Contributor/Innovation Pack/Custom Homepage] Deletion
- Text: Are you certain you want to delete this Space? Keep in mind that this action is irreversible. You can only delete a Space if there are no underlying Subspaces associated with it. If there are any Subspaces within this page, please delete them first (you can do so from the Space settings > Subspaces page).
- Checkbox + text: Please check this box if you are certain about deleting this Space
- Buttons: Cancel; Yes, Delete

### VC
- [ ] Add "more" icon button on VC card
- [ ] This button opens a little menu with -for now only- a delete option
- [ ] Clicking delete opens a confirmation dialog:
- Title: Confirm [Virtual Contributor/Innovation Pack/Custom Homepage] Deletion
- Text: Are you certain you want to delete this [Virtual Contributor/Innovation Pack/Custom Homepage]? Keep in mind that this action is irreversible.
- Checkbox + text: Please check this box if you are certain about deleting this [Virtual Contributor/Innovation Pack/Custom Homepage]
- Buttons: Cancel; Yes, Delete

### Innovation Pack
- [ ] Add "more" icon button on Innovation pack card
- [ ] This button opens a little menu with -for now only- a delete option
- [ ] Clicking delete opens the same confirmation dialog as above (VC)

### Custom Homepage
- [ ] Add "more" icon button on Custom Homepage card
- [ ] This button opens a little menu with -for now only- a delete option
- [ ] Clicking delete opens the same confirmation dialog as above (VC)

## Additional Context

Screen shots / UX designs / data that are relevant.

## Areas that will be affected 
 
_To be added during the refinement_

SimoneZaza:
@bobbykolev Sorry, I've missed your last comment. Probably too late, but:
- Error when deleting space with subspaces is sort of fine, we're warning for this in the confirmation dialog
[...]

PR #6842, opened 2024-09-02
Deletion of entities on Account Tab

bobbykolev:
https://github.com/alkem-io/client-web/issues/6808

- [x] Space Deletion
  - [x] Badge Card was extended to support actions; ⚠️
  - [x] Journey Card Horizontal was extended to support Menu w actions; ⚠️
- [x] VC Deletion
  - [x] Horizontal card extended to support Menu; 
- [x] Packs deletion
  - [x] Inno Pack card extended to support Menu w Actions; 
- [x] Hubs deletion
- [ ] todo: Translate the successful deletion messages;

___
- [x] AccountTabView moved to the domain->community->contributors folder, to be close to the User Settings pages; The ORG settings pages should be there as well; 
- [x] New ActionsMenu component added to the core/ui

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -1416,12 +1416,13 @@
       "visibility": "This Space is currently in <strong>{{visibility}}</strong> mode",
       "readOnly": "If you want to change any of this information, please contact the Alkemio team here."
     },
-    "deleteSpace": {
+    "deleteEntity": {
       "title": "Danger Zone",
       "description": "Click here to delete this {{entity}}. Be careful, this action cannot be undone.",
       "confirmDialog": {
         "title": "Confirm {{entity}} Deletion",
         "description": "Are you certain you want to delete this {{entity}}? Keep in mind that this action is irreversible. You can only delete a {{entity}} if there are no underlying Subspaces associated with it. If there are any Subspaces within this page, please delete them first (you can do so from the {{entity}} settings > Subspaces page).",
+        "descriptionShort": "Are you certain you want to delete this {{entity}}? Keep in mind that this action is irreversible.",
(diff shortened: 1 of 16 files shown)


valeksiev (approved):
I guess it's more for the server but deleting a space is not exactly useful as is now - I got an error saying `Can not delete because the space contains as subspace`. So I have to go the space page, manually delete all subspaces and then delete the space. Maybe we can discuss adding some sort of _force delete_ mode for the mutation when used in the `Account` tab which would cascade delete the space tree.

Not related to this work, but there are a couple of minor UI issues that should be easy fixes:
This misalignment:
image

Also this title displayed in the popups:
image

The UI things are not blocking but I feel we should discuss the functionality for deletion.

Approving as code looks OK.

bobbykolev:
> I guess it's more for the server but deleting a space is not exactly useful as is now - I got an error saying `Can not delete because the space contains as subspace`. So I have to go the space page, manually delete all subspaces and then delete the space. Maybe we can discuss adding some sort of _force delete_ mode for the mutation when used in the `Account` tab which would cascade delete the space tree.
> 
> Not related to this work, but there are a couple of minor UI issues that should be easy fixes: This misalignment: image
> 
> Also this title displayed in the popups: image
> 
> The UI things are not blocking but I feel we should discuss the functionality for deletion.
> 
> Approving as code looks OK.

Thanks @valeksiev 
1. the cascading deletion was discussed and having this 'obstacle' is nice-to-have as someone could delete sub/subspaces of other people. Considering the fact that we're not doing soft deletion this seems ok for now.
2. Yes, we could improve the UI of the entire tab, currently we're reusing existing components and the sizing of the space cards is different;
3. This is mainly due to file service not working locally for you. The title is an SEO alt tag in case of a missing image. However, the tooltip seems redundant.

I'll merge the changes, we can extract improvements separately.

Issue #6843, opened 2024-09-03
authorization policy information page to show additional information


techsmyth:
## Description

As a ops manager I want to be able to track down issues 

## Acceptance criteria

- [ ] Authorization policy page on admin pages of client to show a) type b) the criterias for each rule

## Additional Context

Screen shots / UX designs / data that are relevant.

## Areas that will be affected 
 
_To be added during the refinement_

techsmyth:
```
  lookup {
    authorizationPolicy(ID:"927ac37f-cfca-49c5-8fa7-ee3e2fdccf34") {
      type
      credentialRules {
        grantedPrivileges
        criterias {
          type
          resourceID
        }
      }
    }
  }
```

Comoque1:
Tested OK on DEV env:
<img width="1170" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0hDQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1400ece17d447e9d488000c99e8ab289cf3230dd/image.png" alt="image.png" />

PR #6847, opened 2024-09-05
Criterias added to auth policy info page

bobbykolev:
https://github.com/alkem-io/client-web/issues/6843

- [x] Criteria (type and resourceID) added for every rule;
- [x] Type of the policy;

Screenshot 2024-09-05 143543


techsmyth (approved):
Already a ton better :)

One observation is that if there are multiple criterias then nice to have them on multiple lines, right now they are not all visible:
image

techsmyth (approved):

PR #6862, opened 2024-09-10
initial usage of temp location for file upload

techsmyth:
For now just makes all file uploads temporary so can test it out. Code will need to be updated to have this selectable at the places where the entity is being created.


bobbykolev:
implemented in another PR

PR #7053, opened 2024-10-18
Update web client docs

reactoholic:
Adding info in the docs about `type` vs `interface` and `enum`.

## Summary by CodeRabbit

- **Documentation**
	- Expanded `code-guidelines.md` to include new sections on TypeScript usage.
	- Added insights on the differences between `type` and `interface`.
	- Introduced a section discussing the drawbacks of using `enum` and recommended alternatives.
	- Included best practices and "gotchas" for TypeScript development.

### DIFF
--- a/docs/code-guidelines.md
+++ b/docs/code-guidelines.md
@@ -169,7 +169,7 @@ Usages:
 
 3. If you create a component (`ComponentType`) to pass as a prop, always wrap into `useMemo()` to preserve the component identity. If the component identity changes, React will always remount (that basically disables incremental rendering for the component).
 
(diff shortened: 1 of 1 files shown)


valentinyanakiev (commented):
These are discussion points rather than guidelines, I suggest we discuss with the team and agree on approach and then have it written down.

Plus, we don't need all the details how a decision was made to use something (e.g. text like `even Anders Hejlsberg ...` - for me that is noise in code guidelines file).

reactoholic:
Closing this PR because of not signed commit

PR #7164, opened 2024-11-11
SVG icon as a React component for flexibility (subspace in events view)

bobbykolev:
I want to integrate the SVG icons as React components so we can pass props and easily change their fill (not only).

Please take a look at the following PR, and let's discuss separately how we want to use the custom SVGs in the client.

## Summary by CodeRabbit

- **New Features**
	- Introduced the `SubspaceIcon2` component for improved SVG icon rendering.
	- Enhanced `CardHeaderDetail` component to accept JSX elements for icons, allowing for more flexible usage.
  
- **Bug Fixes**
	- Updated icon imports and usage in `EventCardHeader` for consistent rendering.

- **Documentation**
	- Added guidelines for SVG and icon usage in the documentation to streamline the addition of new icons.

- **Chores**
	- Marked `SubspaceIconV2` as deprecated, advising users to transition to `SubspaceIcon2`.


ccanos:
I have been investigating a bit and read a few articles, this one I liked the most: https://blog.logrocket.com/how-to-use-svgs-react/
Some of them mention react-svg package which I am curious about because in theory it lazy loads the SVG files, but I haven't been able to make it work in 5 minutes so I have left it for another moment (looked like I had to fight with VITE a bit).

If  we need to pass an element with different styles also, `icon` could be just a `ReactNode` instantiated from outside... but we need to decide a single approach for SVGs, we have already a few different ones.

With this approach, I would rename `Icon` to `icon` because it is a prop, and then in the line 22  rename the prop to Icon `icon: Icon,`  in order to use it as `<Icon`, that's what we are doing in other places.

reactoholic (approved):

PR #7213, opened 2024-11-18
remove .env.base from git, the file is generated on build

bobbykolev:
Let's stop tracking the changes on every build in .env.base. Locally the file is created on build. This should be the case in the other environments as well. 

## Summary by CodeRabbit

- **Chores**
	- Removed obsolete environment variable file, streamlining configuration management.
	- Updated `.gitignore` to improve organization and clarity, adding new patterns for ignored files and consolidating similar entries.


hero101 (commented):
run `cross-env NODE_ENV=production node buildConfiguration`
Check if this generates the file with the correct content and you are good to go

Issue #7151, opened 2024-11-06
BUG: Space member, subspace admin, sees button to create subspace template, but he doesn't have the rights


Comoque1:
**Describe the bug**

**To Reproduce**

1. Go to https://acc-alkem.io/ea-publicspace/settings/challenges Space member, subspace admin,
2. Try to create subspace template - error is thrown: `"message": "Authorization: unable to grant 'create' privilege: templatesSet create template from Collaboration, templatesSetId: f5a3b36a-c6ed-4cb5-a121-2c43457b4edb user: 2ef447b5-0b98-4f6e-8bbf-3f3097c61a17 on authorization 05c468dd-b659-4aac-a488-f655d903c45e of type 'templates-set'",`
<img width="1640" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNENjQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e22e53574e457d1f8f499ed3f836e1ea81793cea/image.png" alt="image.png" />

**Expected behavior**
"Create template" is not displayed to subspace admin

Comoque1:
The above fix works on DEV.
There is 1 more location, which should be handled "subspaces" under L1 level in settings. there is a menu button, where we also show "save as template"
<img width="1582" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMmZpQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4f8c75d112631198814c7329aad5e123c9e29480/image.png" alt="image.png" />
<img width="1222" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMmppQnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f11849334003ac3d58ddcdb3006512d6de969e13/image.png" alt="image.png" />

PR #7232, opened 2024-11-20
Create Template in Subspace settings - check for parent space permissions

bobbykolev:
https://github.com/alkem-io/client-web/issues/7151

To solve the issue I used the **Update** permission on the parent space.
The spaceId was not available so I passed it from where we're gathering the journeyId.
There are two separate button components based on the premission because in a disabled state the Tooltip is not triggered.

## Summary by CodeRabbit

- **New Features**
	- Added community engagement sections and updated footer links for improved navigation.
	- Enhanced template management in Space Settings, including user privilege checks for creating templates.
	- Improved accessibility with updated alt text for visuals and additional validation messages in forms.

- **Bug Fixes**
	- Refined user-facing messages in authentication and error display sections for clarity.

- **Documentation**
	- Updated strings and sections in the translation files to enhance user experience across the platform.


techsmyth:
@bobbykolev another thought: every space can access the templatesManager & then the templates set in that. So possible to check for create directly on the templatesSet? Would also avoid need to pass around parent space id…

bobbykolev:
> @bobbykolev another thought: every space can access the templatesManager & then the templates set in that. So possible to check for create directly on the templatesSet? Would also avoid need to pass around parent space id…

@techsmyth , I can't quite understand your suggestion. 
We are trying to create the template in the templateSet and the fix is to catch it before trying to. Is there another privilege or you are suggesting something else?

techsmyth:
If you are planning to create the template, then you need to have the templatesSet ID. So why not just do a check for the CREATE privilege on the TemplatesSet entity? And second point is that you can easily get to the templatesSet directly from any space via the TemplatesManager.

bobbykolev:
> If you are planning to create the template, then you need to have the templatesSet ID. So why not just do a check for the CREATE privilege on the TemplatesSet entity? And second point is that you can easily get to the templatesSet directly from any space via the TemplatesManager.

Got it. It makes sense to check that instead of 'Update' on space, thanks!

ccanos (approved):

PR #7254, opened 2024-11-25
Notifications v0.1

bobbykolev:


bobbykolev:
@coderabbitai review

bobbykolev:
@coderabbitai configuration

src/main/inAppNotifications/InAppNotificationsDialog.tsx
```diff
@@ -0,0 +1,21 @@
...
+import DialogWithGrid from '../../core/ui/dialog/DialogWithGrid';
+import DialogHeader from '../../core/ui/dialog/DialogHeader';
+import { useInAppNotificationsContext } from './InAppNotificationsContext';
+import { InAppNotificationsList } from './InAppNotificationsList';
+
+export const InAppNotificationsDialog = () => {
+  const { t } = useTranslation();
+  const { isOpen, setIsOpen } = useInAppNotificationsContext();
+
+  return (
```
ccanos, line 13: We are not doing it often, but I think in this case that this component sits very high and will be rendered in every page, it would be good if we just return `null` if is not open to avoid rendering unnecessary things. 
This dialog doesn't have any state that we want to keep when closed and I guess that was your intention already with that `{isOpen && <InAppNotificationsList />}` to avoid calling the query?

src/main/inAppNotifications/useInAppNotifications.ts
```diff
@@ -0,0 +1,133 @@
...
+  comment?: string;
+  commentUrl?: string;
+}
+
+export const useInAppNotifications = () => {
+  const { user, platformRoles } = useUserContext();
+
+  const [updateState] = useUpdateNotificationStateMutation();
+
+  const enableNotifications = useMemo(() => {
```
ccanos, line 95: I think this should go into the `InAppNotificationsContext`. 
It may be a bigger refactor than I think, but this context sits under the UserContext so it should have no problem to access these, and this way we'll unify the place where we change if notifications are enabled for this user or not... Does that make sense? disregard if it's too big refactor, I've only seen this check twice for now

src/main/inAppNotifications/graphql/InAppNotificationsFragments.graphql
```diff
@@ -0,0 +1,95 @@
...
+      displayName,
+      url,
+      visual(type: CARD) {
+        ...VisualUri
+      }
+    }
+  }
+  actor {
+    id
+    __typename
```
ccanos, line 68: add a comment smiliar to the one in `src/domain/journey/space/pages/SpaceAccount/SpaceAccount.graphql` please, in theory it should always be there but wanted to keep track of the `__typename`s that we actually use

ccanos (changes requested):

src/core/ui/icon/BadgeCounter.tsx
```diff
@@ -11,7 +11,7 @@ const BadgeSizes: Record<NonNullable<BadgeCounterProps['size']>, TypographyProps
   small: {
     fontSize: '0.7rem',
     lineHeight: gutters(0.8),
-    width: gutters(0.8),
+    minWidth: gutters(0.8),
```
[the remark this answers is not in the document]
bobbykolev: Currently, if you have 10, or 33, or more notifications, the number is cropped as you have a strict circle.
With this change, now if the content expands beyond a single digit, the badge becomes a rectangle instead of circle (rounded).

src/main/inAppNotifications/InAppNotificationsDialog.tsx
bobbykolev: Yes, indeed. 
If we return null, potentially the closing of the dialog will be instant and any animation (we don't have currently) will suffer. Do you think it would make a difference if we render the empty dialog? 
@ccanos

src/main/inAppNotifications/useInAppNotifications.ts
bobbykolev: It doesn't require big refactoring. Let me think about it conceptually as on first thought, I'd like to keep the context as independent as possible.

bobbykolev: I rethought it and I agree that this suggestion is the right thing to do, applying separation of concerns, consistency, and DRY. Performance-wise it's not ideal, but it's not something critical either. Thanks, changes applied, please double-check.

src/main/inAppNotifications/InAppNotificationsDialog.tsx
ccanos: nah, leave it. You're right I didn't think about animations, good point. I don't think a closed dialog can be heavy at all. Agreed

ccanos (approved):

PR #7258, opened 2024-11-26
refactor support dialog

reactoholic:
## Refactor support dialog

### DIFF
--- a/src/core/help/dialog/HelpDialog.tsx
+++ b/src/core/help/dialog/HelpDialog.tsx
@@ -1,13 +1,15 @@
-import { useTranslation } from 'react-i18next';
+import { Trans, useTranslation } from 'react-i18next';
 import { Box, Dialog, Grid, Link, styled } from '@mui/material';
-import QuizOutlinedIzon from '@mui/icons-material/QuizOutlined';
+import QuizOutlinedIcon from '@mui/icons-material/QuizOutlined';
 import ForumOutlinedIcon from '@mui/icons-material/ForumOutlined';
 import FiberNewTwoToneIcon from '@mui/icons-material/FiberNewTwoTone';
-import ArrowForwardIcon from '@mui/icons-material/ArrowForward';
 import { DialogContent } from '@/core/ui/dialog/deprecated';
 import WrapperMarkdown from '@/core/ui/markdown/WrapperMarkdown';
 import { useConfig } from '@/domain/platform/config/useConfig';
 import DialogHeader from '@/core/ui/dialog/DialogHeader';
+import { gutters } from '@/core/ui/grid/utils';
+import { TopLevelRoutePath } from '@/main/routing/TopLevelRoutePath';
+import useServerMetadata from '@/domain/platform/metadata/useServerMetadata';
 
 interface HelpDialogProps {
   open: boolean;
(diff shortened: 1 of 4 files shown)


src/core/help/dialog/HelpDialog.tsx
```diff
@@ -53,41 +46,77 @@ const CustomNewIcon = styled(FiberNewTwoToneIcon)(() => ({
...
-  const contactUs = `${locations?.support}`;
-  const gettingStarted = `${locations?.help}/getting-started`;
+
+  const supportHref = locations?.support;
+  const docsHref = `/${TopLevelRoutePath.Docs}`;
+  const welcomeSpaceHref = `/${TopLevelRoutePath.WelcomeSpace}`;
 
   return (
-    <Dialog open={open} onClose={handleClose} maxWidth={'xl'}>
+    <Dialog open={open} onClose={handleClose} sx={{ '& .MuiDialog-paper': { maxWidth: 700 } }}>
```
bobbykolev, line 61: Please, use DialogWithGrid and pass columns.

src/core/help/dialog/HelpDialog.tsx
```diff
@@ -53,41 +46,77 @@ const CustomNewIcon = styled(FiberNewTwoToneIcon)(() => ({
...
+              {t('pages.help-dialog.icons.joinTheWelcomeSpace')}
             </IconWrapper>
           </Grid>
         </Grid>
-        <FooterLink href={locations?.tips} target="_blank">
-          <ArrowForwardIcon />
-          {t('pages.help-dialog.tips-and-tricks')}
-        </FooterLink>
+
+        <Box sx={{ display: 'flex', justifyContent: 'center', marginTop: gutters(2) }}>
```
bobbykolev, line 103: Information: We tend to use our wrappers over the default MUI components. So we can either use Grid, Gutters, Actions, etc. instead of Box.

In this case you can use <Gutters row> wich automatically will give gap to the items below, no need to define margins.

src/core/help/dialog/HelpDialog.tsx
```diff
@@ -53,41 +46,77 @@ const CustomNewIcon = styled(FiberNewTwoToneIcon)(() => ({
...
+          <Box sx={{ marginLeft: gutters(2) }}>
+            <Trans
```
bobbykolev, line 105: Please, put the text in a Caption. As per design, this information shouldn't be focused.

src/core/help/dialog/HelpDialog.tsx
```diff
@@ -22,28 +24,19 @@ const HelpDialogContent = styled(DialogContent)(({ theme }) => ({
...
   display: 'block',
   position: 'relative',
-  padding: theme.spacing(4),
+  padding: theme.spacing(2),
   textAlign: 'center',
   fontWeight: 'bold',
 }));
 
 const Icon = styled(Box)(({ theme }) => ({
   color: theme.palette.primary.dark,
```
bobbykolev, line 33: Looking into the design here should be primary.main

bobbykolev: Same on line 20.

bobbykolev (commented):
A couple of small items, it shouldn't take you more than 15 minutes, and then we can merge.

src/core/help/dialog/HelpDialog.tsx
reactoholic: > Please, put the text in a Caption. As per design, this information shouldn't be focused.

The versions themselves are bolded in the design. Should I replace the `<b>` tag with `<caption>` tag?

bobbykolev (approved):

PR #7298, opened 2024-12-10
Remove VC entitlement check for adding VC to community

valentinyanakiev:
- removed check of entitlements for adding VCs to community
- this is error prone, as the `isAvailable` field on the entitlement is checked, and that returns `false` when we run out of slots
- regardless, the check is on the `User` `Account` and that may be wrong, because the VC may be coming from `Org` `Account` - so that requires discussion...


Comoque1:
Tested OK locally.

PR #7356, opened 2024-12-18
[VC store] the new flow from settings

bobbykolev:
Now the _(Sub)(Sub)Space -> Settings -> Community_ uses the new VC Store dialog.

- The functionality of adding/removing/inviting works as expected.
- The logic for gathering/filtering VCs was further simplified (it was too complex). 
- There's one button - Add - similar to the other tables in the Layout (the invite one was removed as both are showin the same dialog).

However, there are a couple of **drawbacks**:
- In the new flow there's **no search** ⚠️
- The _On Account_ list shows **all** VC under the account. On SubSpace level this could cause an error on attempt to ad a VC not part of the parent community.

### DIFF
--- a/src/domain/community/inviteContributors/InviteVCsDialog.tsx
+++ b/src/domain/community/inviteContributors/InviteVCsDialog.tsx
@@ -20,12 +20,20 @@ import { useRouteResolver } from '@/main/routing/resolvers/RouteResolver';
 import PageContentBlockHeader from '@/core/ui/content/PageContentBlockHeader';
 import { gutters } from '@/core/ui/grid/utils';
 import { Caption } from '@/core/ui/typography';
+import { useSubSpace } from '@/domain/journey/subspace/hooks/useSubSpace';
+import { useOpportunity } from '@/domain/journey/opportunity/hooks/useOpportunity';
 
 const InviteVCsDialog = ({ open, onClose }: InviteContributorDialogProps) => {
   const { t } = useTranslation();
   const notify = useNotification();
 
-  const { spaceId, roleSetId } = useSpace();
+  const { opportunityId: subSubSpaceId, roleSetId: oppRoleSetId } = useOpportunity();
+  const { subspaceId: subSpaceId, roleSetId: subSpaceRoleSetId } = useSubSpace();
+  const { spaceId: spaceSpaceId, roleSetId: spaceRoleSetId } = useSpace();
+
+  const spaceId = subSubSpaceId || subSpaceId || spaceSpaceId;
+  const roleSetId = oppRoleSetId || subSpaceRoleSetId || spaceRoleSetId;
+
   const { spaceLevel } = useRouteResolver();
 
   // data
@@ -52,7 +60,7 @@ const InviteVCsDialog = ({ open, onClose }: InviteContributorDialogProps) => {
   const [bokProfile, setBoKProfile] = useState<BasicSpaceProps>();
 
   const fetchVCs = async () => {
-    let acc = await getAvailableVirtualContributors('', false);
+    let acc = await getAvailableVirtualContributors('');
     setOnAccount(acc);
 
     let lib = await getAvailableVirtualContributorsInLibrary('');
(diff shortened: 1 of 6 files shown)


bobbykolev:
@reactoholic , this one is tricky. The _InviteVCsDialog_ now has all the logic related to understanding the (sub)space context (previously working only for the `spaceLevel === Space`). Based on that, it gathers the required data through the _useInviteContributors_ hook. The filtering logic for displaying the VCs based on the ones in Account, already in the community, and from the library is shared between the two components. 

Ideally, we should:
- add search (as it was), it should be part of another story;
- filter the VCs on Account by existence in the parent community (the tricky part);

As part of the PR, please test the invites/adding from the Account through the Community Settings for all spaceLevels.

src/domain/community/inviteContributors/InviteVCsDialog.tsx
```diff
@@ -52,7 +60,7 @@ const InviteVCsDialog = ({ open, onClose }: InviteContributorDialogProps) => {
   const [bokProfile, setBoKProfile] = useState<BasicSpaceProps>();
 
   const fetchVCs = async () => {
-    let acc = await getAvailableVirtualContributors('', false);
+    let acc = await getAvailableVirtualContributors('');
```
reactoholic, line 63: Let's convert these two variables to `const` since we're not reassigning new values to them.
Also, I see in my editor we don't have a lint rule for using `let` when not reassigning a new value to it. I suggest we add one.

src/domain/community/inviteContributors/InviteVCsDialog.tsx
```diff
@@ -109,6 +117,11 @@ const InviteVCsDialog = ({ open, onClose }: InviteContributorDialogProps) => {
     return onAccount?.find(c => c.id === id) || inLibrary?.find(c => c.id === id);
   };
 
+  // remove the duplicates from available onAccount in the inLibrary list
```
reactoholic, line 120: Add a `TODO:` in front of this message.

src/domain/community/inviteContributors/InviteVCsDialog.tsx
```diff
@@ -109,6 +117,11 @@ const InviteVCsDialog = ({ open, onClose }: InviteContributorDialogProps) => {
...
+  const filteredInLibrary = useCallback(() => {
```
reactoholic, line 121: While this is perfectly working implementation of this method, I think we can lower down from `O(n * m)` to `O(n + m)` (`n` is `inLibrary`'s length, `m` is `onAccount`'s length) by doing so:
```
const filteredInLibrary = useCallback(() => {
  if (!inLibrary) return [];

  // Creating a Map for quick access to elements in onAccount by id. This allows the check if id is present to be lowered down to O(1), e.g., reduces execution time from quadratic to linear due to the use of a hash table (Map).
  const accountIds = new Map(onAccount?.map(member => [member.id, true]));

  // inLibrary filtering where we check if id is present in accountIds
  return inLibrary.filter(vc => !accountIds.has(vc.id));
}, [inLibrary, onAccount]);
```

src/domain/community/inviteContributors/useInviteContributors.tsx
```diff
@@ -110,35 +109,25 @@ const useInviteContributors = ({
   };
 
   const [fetchAllVirtualContributors, { loading: accountVCsLoading }] = useAvailableVirtualContributorsLazyQuery();
-  const getAvailableVirtualContributors = async (filter: string | undefined, all: boolean = false) => {
+  const getAvailableVirtualContributors = async (filter: string | undefined) => {
     const { data } = await fetchAllVirtualContributors({
```
reactoholic, line 113: I suggest wrapping this in `try/catch` since we're using `async` here.

reactoholic (changes requested):

src/domain/community/inviteContributors/InviteVCsDialog.tsx
bobbykolev: this is explanation not a TODO

src/domain/community/inviteContributors/InviteVCsDialog.tsx
bobbykolev: We can do it. I used to think more about that in the past, but this syntaxis sugar is mainly used in this project.
The VCs on the account are not more then 3 but the library ones could grow so it makes sense, thanks.

src/domain/community/inviteContributors/useInviteContributors.tsx
bobbykolev: Hm, we have the global handler which would notify in case of an error. Do you have anything specific in mind?

src/domain/community/inviteContributors/InviteVCsDialog.tsx
bobbykolev: I've implemented it with Set and extracted it as a util for getting the unique items, we're using such a comparison on many places.

reactoholic (approved):
Work very good

bobbykolev:
I'm closing for now because it needs more refinement and work before replacing the existing functionality.

PR #7387, opened 2025-01-03
implement image pasting in md editor

reactoholic:
implement image pasting in md editor

## Summary by CodeRabbit

- **New Features**
  - Enhanced markdown input field with image paste support.
  - Improved character count validation with visual feedback for overflow.
  - Updated event handling and prop management for markdown input components.
  - Integrated user notifications for successful file uploads.

- **Documentation**
  - Added clarifying comments for URL protocol handling.

The changes focus on improving the markdown input experience with advanced editing capabilities and better file upload support.

### DIFF
--- a/src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
+++ b/src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
@@ -150,31 +150,24 @@ export const FormikMarkdownField = ({
         for (const item of items) {
           if (item.type.startsWith('image/')) {
             const file = item.getAsFile();
-
             if (file) {
-              for (const item of items) {
-                if (item.type.startsWith('image/')) {
-                  const file = item.getAsFile();
-                  if (file) {
-                    const reader = new FileReader();
-                    reader.onload = () => {
-                      uploadFile({
-                        variables: {
-                          file,
-                          uploadData: {
-                            storageBucketId,
-                            temporaryLocation: true,
-                          },
-                        },
-                      });
-                    };
-                    reader.readAsDataURL(file);
-                  }
-                  event.preventDefault();
-                }
-              }
+              const reader = new FileReader();
+
+              reader.onload = () => {
+                uploadFile({
+                  variables: {
+                    file,
+                    uploadData: {
+                      storageBucketId,
+                      temporaryLocation: true,
+                    },
+                  },
+                });
+              };
+
+              reader.readAsDataURL(file);
+              event.preventDefault();
             }
-            event.preventDefault();
           }
         }
       }
--- a/src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
+++ b/src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
@@ -1,6 +1,25 @@
-import { ChangeEvent, useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
-import { FormControl, FormHelperText, InputLabel, InputProps, OutlinedInput, useFormControl } from '@mui/material';
+import {
+  ChangeEvent,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+  ClipboardEvent,
+  PropsWithChildren,
+} from 'react';
+import {
+  FormControl,
+  FormHelperText,
+  InputBaseComponentProps,
+  InputLabel,
+  InputProps,
+  OutlinedInput,
+  useFormControl,
+} from '@mui/material';
 import { useField } from 'formik';
+import { Editor } from '@tiptap/react';
 import CharacterCounter from '../characterCounter/CharacterCounter';
 import TranslationKey from '@/core/i18n/utils/TranslationKey';
 import { useValidationMessageTranslation } from '@/domain/shared/i18n/ValidationMessageTranslation';
@@ -11,6 +30,9 @@ import { MarkdownTextMaxLength } from '../field-length.constants';
 import { error as logError } from '@/core/logging/sentry/log';
 import { isMarkdownMaxLengthError } from './MarkdownValidator';
 import { useTranslation } from 'react-i18next';
+import { useUploadFileMutation } from '@/core/apollo/generated/apollo-hooks';
+import { useNotification } from '../../notifications/useNotification';
+import { useStorageConfigContext } from '@/domain/storage/StorageBucket/StorageConfigContext';
 
 interface MarkdownFieldProps extends InputProps {
   title: string;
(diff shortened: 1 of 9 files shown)


src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
```diff
@@ -97,6 +135,46 @@ export const FormikMarkdownField = ({
...
+            if (file) {
+              const reader = new FileReader();
+
+              reader.onload = () => {
+                uploadFile({
+                  variables: {
+                    file,
+                    uploadData: {
+                      storageBucketId,
+                      temporaryLocation: true,
```
bobbykolev, line 162: Maybe storageConfig.temporaryLocation, is always temporary?

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
```diff
@@ -133,28 +219,27 @@ export const FormikMarkdownField = ({
             {title}
           </InputLabel>
         )}
+
         <OutlinedInput
-          value={field.value}
-          onChange={handleChange}
-          onBlur={handleBlur}
+          multiline
```
bobbykolev, line 224: It's not critical but why all the props are reordered? It makes it harder to spot the actual changes.

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
```diff
@@ -97,6 +135,46 @@ export const FormikMarkdownField = ({
...
+                    file,
+                    uploadData: {
+                      storageBucketId,
+                      temporaryLocation: true,
+                    },
+                  },
+                });
+              };
+
+              reader.readAsDataURL(file);
```
bobbykolev, line 168: 1. Do we need this `reader.readAsDataURL(file);`?

2. Should we preventDefault earlier?

bobbykolev (commented):
It works very well, I like the feature of pasting multiple images 👏

I left a few comments, nothing critical but double-check the temporary logic. 
This passing of the editor looks odd to me, but I can't suggest something better.

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
reactoholic: It's temporary just on initialization.

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
reactoholic: 1. Added.
2. The order of calling `event.preventDefault()` in the code is correct because it is invoked after verifying that the clipboard content is an image and before the event is propagated to the rest of the application. This prevents the browser from performing its default paste behavior (e.g., adding the image to the editor content by default), but I found even a more optimized way to call `event.preventDefault()` by checking if there's an actual `image` -> only then call `event.preventDefault()`.

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
reactoholic: Sorry for that, I reordered them for me to be easy to orient and forgot to move them back to their original order after I finished with it.

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
bobbykolev: As far as I'm aware the logic should be - if you don't have the entityId, then save the doc as temporary. If not, you should use the existing entityId. 
@ccanos do you think we need to handle this, or we're fine to use temp always and as Petar said - the server is making sure to move all files to their storage after they are used and we can always safely use temp storage?

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
```diff
@@ -70,8 +94,25 @@ export const FormikMarkdownField = ({
...
   const [field, meta, helper] = useField({ name, validate });
 
+  const [uploadFile] = useUploadFileMutation({
+    onCompleted: data => {
+      notify(t('components.file-upload.file-upload-success'), 'success');
+
+      editor?.commands.setImage({ src: data.uploadFileOnStorageBucket, alt: 'pasted-image' });
+    },
+    onError: error => {
+      logError(error);
```
bobbykolev, line 108: Please double-check that we're not logging this error twice. If the uploading fails I guess we have the global handler.

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
```diff
@@ -140,13 +140,13 @@ export const FormikMarkdownField = ({
       const clipboardData = event.clipboardData;
       const items = clipboardData.items;
 
-      if (!items) {
-        return;
-      }
+      if (!items) return;
```
bobbykolev, line 146: I like this shortening, but I think we agreed that we're not using it. @ccanos to confirm

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
bobbykolev: network

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
reactoholic: > As far as I'm aware the logic should be - if you don't have the entityId, then save the doc as temporary. If not, you should use the existing entityId. @ccanos do you think we need to handle this, or we're fine to use temp always and as Petar said - the server is making sure to move all files to their storage after they are used and we can always safely use temp storage?

The `temporaryLocation` flag is handled from the parent so no logic regarding it is required here.

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
```diff
@@ -72,6 +70,10 @@ export const FormikMarkdownField = ({
 
   const [field, meta, helper] = useField({ name, validate });
 
+  const { t } = useTranslation();
```
bobbykolev, line 73: Please revert this file to its original. There are no actual changes here related to the task.

src/core/ui/forms/MarkdownInputControls/MarkdownInputControls.tsx
```diff
@@ -43,6 +43,7 @@ const sanitizeUrl = (url: string): string => {
     const parsedUrl = new URL(url);
     const allowedProtocols = ['http:', 'https:'];
     // 'javascript:' is used to prevent XSS attacks by blocking dangerous protocols
+    // eslint-disable-next-line no-script-url
```
bobbykolev, line 46: no need to change this file in this PR as well

src/core/ui/forms/MarkdownInput/FormikMarkdownField.tsx
reactoholic: Correct, I have moved all the new things here to another component.

bobbykolev (approved):
Let's remove the redundant changes. 
Otherwise, the pasting is handled and works very well 👏

One small thing that could be extracted as an enhancement is a loading indication. If you paste a large file under a slower network, nothing happens for seconds.

bobbykolev (approved):

PR #7389, opened 2025-01-06
common Alkemio terms added to the dictionary

bobbykolev:
add domain-specific words to dict


bobbykolev:
If this is not working or it takes too much time, please leave it @reactoholic 
It should have been a quick update of the dict without the need of further changes.

PR #7391, opened 2025-01-06
InnovationFlow States names validation

ccanos:
- Avoiding commas should be enough, but in case we end up serializing tags as JSON string arrays, I have forbidden some JS special characters too that should be fine to live without: `[ ] ' " , { } \`

- Also added a Gutters for better visualization of the form.


bobbykolev (dismissed):

src/domain/collaboration/InnovationFlow/InnovationFlowDragNDropEditor/InnovationFlowStateForm.tsx
```diff
@@ -41,6 +42,10 @@ const InnovationFlowStateForm = ({
       .string()
       .required()
       .max(SMALL_TEXT_LENGTH)
+      // Avoid any JSON-related chars:
+      // This validation is also performed on the server: domain/collaboration/innovation-flow-states/innovation.flow.state.service.ts
+      // Keep them in sync
+      .matches(/^[^[\],'"{}\\]+$/, t('components.innovationFlowSettings.stateEditor.invalidChars'))
```
bobbykolev, line 48: it's more performant to define the pattern first and then use it in match:
const pattern = /^[^[\],'"{}\\]+$/;
.match(pattern);

bobbykolev (approved):

PR #7428, opened 2025-01-10
Client web/7416 disable image pasting when hide image options flag is true

reactoholic:
## Summary by CodeRabbit

- **New Features**
  - Enhanced markdown input paste handling with improved checks for image and HTML content.
  - Added option to prevent pasting images when the hide image options feature is enabled.

- **Style**
  - Updated comments for improved clarity in the markdown input component.
  - Modified `BlockTitle` component styling in the organization verified status view.


bobbykolev (commented):
@reactoholic I still think you're doing two redundant for of loops. Also doubling the check for image/.
It's good that you defined a method for checking and I'd suggest:
- extract the function ouside of the EditorOptions;
- insted of extracting the check, extract the onImagePaste logic, and place the check for hideImageOptions, in a single for of loop.

reactoholic:
> @reactoholic I still think you're doing two redundant for-of loops. Also doubling the check for image/. It's good that you defined a method for checking and I'd suggest:
> 
> * extract the function outside of the EditorOptions;
> * instead of extracting the check, extract the onImagePaste logic, and place the check for hideImageOptions, in a single loop.

Narrowed down to one loop and pushed :)

bobbykolev (approved):
Image pasting is disabled if hideImageOptins is true ✅
Documentation of the paste handle + optimal loop/check logic in place ✅

PR #7443, opened 2025-01-14
activity blocks consistent size in simple home page

reactoholic:
activity blocks consistent size in simple home page

## Summary by CodeRabbit

- **New Features**
  - Reduced the number of displayed contributions from 20 to 10.
  - Added a "Show More" option to view additional activities through a dialog.
  - Introduced a more flexible rendering mechanism for contributions.

- **Improvements**
  - Enhanced user interaction for viewing contributions.
  - Simplified space selection and activity rendering logic.
  - Improved component readability and maintainability.

- **User Experience**
  - More compact initial view of contributions.
  - Easy access to additional activities via a new dialog interface.


reactoholic:
The Dialog render all the activities. A subject to discussion is whether or not we want the filters to be visible in the Dialog as well, because now if the user wants to add filters they must close the Dialog first, apply the filters and re-open the Dialog. If so - this is a subject for a new Issue in ZenHub.

reactoholic:
Converting this PR into **Draft** by Bobbi's request.

src/main/topLevelPages/myDashboard/latestContributions/LatestContributions.tsx
```diff
@@ -150,6 +152,8 @@ const LatestContributions = ({ spaceMemberships }: LatestContributionsProps) =>
     [data?.activityFeed.activityFeed]
   );
 
+  const handleOpenActivitiesDialog = (dialogType: DashboardDialog) => () => setIsOpen(dialogType);
```
bobbykolev, line 155: Let's use the abstraction here to memoize it with useCallback. Otherwise we could've pass setIsOpen directly.

bobbykolev (commented):
1. Blocking Issue - the pagination is not working in the activity dialog accessed through the Spaces Dashboard View;

I can't understand why you're updating the Contribution components specifically their UI logic.
A better approach is to keep these components independent from where they are rendered - Dash block, dialog, or something else. 

I'd suggest:
1. support only passing a limit to the components;
2. implement a simple show more button outside of these components;
3. reuse the DashboardDialogs component in the Activity Dash;

Let me know if you have any questions or other suggestions. I've already discussed that we have the Dialogs in the Space view, so please don't doubled them. @reactoholic

PR #7486, opened 2025-01-22
Account storage config for the account resources

bobbykolev:
We are introducing Account Storage in the client. 

The logic alters from either a storageBucket from the profile or PlatformBucket to:
`profile StorageBucket ?? account directStorageBucket ?? platform's directStorageBucket`

The new storage bucket is used on the Contributors' Account Settings Tab for VCs, Packs and Hubs.  

## Summary by CodeRabbit

- **New Features**
	- Added account-specific storage configuration queries and hooks.
	- Introduced a new GraphQL query `AccountStorageConfig`.

- **Refactor**
	- Replaced `userId` with `accountId` in multiple components.
	- Updated storage configuration context to use account-level identifiers.
	- Modified storage configuration retrieval logic to support account-based configurations.


src/domain/storage/StorageBucket/StorageConfig.graphql
```diff
@@ -120,6 +120,28 @@ query PlatformStorageConfig {
   }
 }
 
+query AccountStorageConfig($accountId: UUID_NAMEID!) {
+  account(ID: $accountId) {
```
ccanos, line 124: how is it possible that $accountId is a `UUID_NAMEID` ?
This is probably a mistake, I'll check in the server, but even if it is wrong in the server, I think it would be better if we access the account through the lookup
`query { lookup { account(ID ...` this is for sure a UUID

ccanos (changes requested):

src/domain/storage/StorageBucket/StorageConfig.graphql
bobbykolev: Yes, I was also surprised but I didn't want to extend this fix. It's in the release bucket, this could be taken separately.

src/domain/storage/StorageBucket/StorageConfig.graphql
```diff
@@ -120,6 +120,28 @@ query PlatformStorageConfig {
...
+        allowedMimeTypes
+        maxFileSize
+        authorization {
+          id
+          myPrivileges
+        }
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
bobbykolev: @ccanos can you double-check. Most of the entities are using profile, but we don't have/use profile on Account. That's why I used the `directStorageBucket` logic. Let me know if this is correct 🙏

src/domain/storage/StorageBucket/StorageConfig.graphql
techsmyth: Account does not actually have a nameID, so this is I think a pure bug. It may be that I fixed the definition of this query in the RoleSet v2 PRs, so check after that is merged.

src/domain/storage/StorageBucket/StorageConfig.graphql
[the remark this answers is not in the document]
ccanos: yes, your code is correct, 
@coderabbit is hallucinating here
there is no profile inside account

ccanos (approved):

PR #7490, opened 2025-01-22
VC Flow - proper filtering of Existing Spaces 

bobbykolev:
- now we filter the spaces before getting their subspaces (causing bugs for org accounts); _the second commit_
- the VC flow components were organized + the 'New' word was removed; _the first commit_

## Summary by CodeRabbit

- **Refactoring**
  - Renamed virtual contributor wizard hook and related components from "New" to standard naming.
  - Updated import paths and references across multiple files.
  - Modified GraphQL query to focus on space subspaces instead of account spaces.

- **Improvements**
  - Streamlined space selection logic in the virtual contributor wizard.
  - Enhanced loading state management for existing spaces.
  - Simplified component props and state handling.

- **Technical Changes**
  - Removed deprecated cache utility functions.
  - Updated component and interface names for consistency.


src/main/topLevelPages/myDashboard/newVirtualContributorWizard/ExistingSpace.tsx
```diff
@@ -81,15 +59,20 @@ const ExistingSpace = ({ onClose, onBack, onSubmit, accountId, getSpaces, loadin
...
   });
 
-  const onCreate = (values: { subspaceId: string }) => {
+  const onCreate = async (values: { subspaceId: string }) => {
+    setSubmitLoading(true);
+
     const bok = listItems.filter(s => s.id === values.subspaceId)[0];
-    bok && onSubmit(bok);
+    if (bok) {
+      await onSubmit(bok);
```
reactoholic, line 73: `onSubmit(bok)` doesn't return a **Promise**, we can remove the `await` here. If in reality `onSubmit(bok)` does return a **Promise** we should update the type instead.

reactoholic (changes requested):

src/main/topLevelPages/myDashboard/newVirtualContributorWizard/ExistingSpace.tsx
bobbykolev: You are right, it's the type as we need to await it in order the loading logic to work.

reactoholic (approved):

PR #7565, opened 2025-01-29
VC store refinements (list page)

reactoholic:
update styles and add search bar

## Summary by CodeRabbit

- **New Features**
	- Added search functionality for Virtual Contributors.
	- Introduced dynamic typography variants for page headers.
	- Added new dialog title for "Virtual Contributors Store."

- **Improvements**
	- Enhanced dialog titles for Virtual Contributors.
	- Updated translation for Virtual Contributors management.
	- Improved filtering of Virtual Contributors in the dialog.

- **User Interface**
	- Added search field for Virtual Contributors.
	- Refined page content block header styling options.


src/domain/community/inviteContributors/InviteVCsDialog.tsx
```diff
@@ -68,6 +75,14 @@ const InviteVCsDialog = ({ open, onClose }: InviteContributorDialogProps) => {
     getAvailableVirtualContributorsInLibrary,
   ]);
 
+  const filterVCs = (virtualContributor: VirtualContributorProps) =>
```
bobbykolev, line 78: I don't think we need this logic here. You just need to pass the filter state value on line 66 and 69 as a first argument of the getAvailableVirtualContributors and getAvailableVirtualContributorsInLibrary

They support the filter logic.

src/domain/community/inviteContributors/InviteVCsDialog.tsx
```diff
@@ -68,6 +75,14 @@ const InviteVCsDialog = ({ open, onClose }: InviteContributorDialogProps) => {
...
+    virtualContributor.profile.displayName.toLowerCase().includes(filter.toLowerCase());
+
+  const filteredVCs = useMemo(
```
bobbykolev, line 81: The first comment will fix this issue but this implementation makes the filtering only for the VCs under the global library. It doesn't apply to the ones for the Account.

bobbykolev:
@reactoholic please review the fix in this commit: https://github.com/alkem-io/client-web/pull/7565/commits/eb407937437e68e7c6185805421b13885d76d232

If you agree with it we can merge.

There's another but related to the latest RoleSet implementation where the initially loaded VCs on the account are not filtered by the already added to the community ones. This might be related to the issue that we have with the updates to the community. Let's tackle it separately.

bobbykolev (approved):

Issue #7591, opened 2025-01-31
[Client Optimizations] Spike - Optimize UserProvider


bobbykolev:
## Description

As a user,
I want a faster application.
As part of the client performance optimizations, let's review and improve the implementation and usage of the UserProvider.

Ideally, we should query only the required data and build up once additional is required - to be defined after the Spike.

## Acceptance criteria

- Should optimize the userProvider query and Client Context implementation
  - [ ] request less data on app load;
  - [ ] reduce the loading time;
  - [ ] ...

## Additional Context

Screen shots / UX designs / data that are relevant.

## Areas that will be affected

_To be added during the refinement_

reactoholic:
## Suggestions:

- Remove the user creation from the provider.
- Remove the fetch of entitlements, since they are used only in useCreateSpaceLink and in all of the 4 places where useCreateSpaceLink is imported entitlements are not being used.
- Extract the account data into a separate file (preferably a hook) and use it on requirement only since the data from useAuthenticationContext is used just once in places the user may not even go so this will exclude additional HTTP calls.

reactoholic:
This spike includes also the actual UserProvider optimization

reactoholic:
## Affected Areas:

**Application-wide**: The impact is application-wide, particularly noticeable during login, registration, and browser hard-refreshes, as the UserProvider component wraps the entire application.

## QA Notes:

**Data Integrity**: There are no changes to the data fetched from the database. The data retrieved remains consistent with the previous implementation.

**Performance Optimization**: The client-side optimizations prevent unnecessary data re-fetching, data calculations, and redundant user creation mutation calls during registration. This results in faster execution of the UserProvider and an improved user experience due to perceived responsiveness.

CC: @Comoque1

Comoque1:
After the changes on ACC:
user: comoque+7010@gmail.com
<img width="1555" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzc3Q0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f8546798da339b1fd310b768b7c68e34f165658b/image.png" alt="image.png" />

User comoque+1011@gmail.om
[...]

PR #7606, opened 2025-02-04
UserProvider Optimization

reactoholic:
#7591 [SPIKE] ~ USER PROVIDER OPTIMIZATION

## Key Improvements and Explanations:

- Prevent Duplicate createUserProfile Calls: Added hasCreatedUser state to ensure the createUserProfile mutation is only called once per user session when the user is first created. This prevents redundant API calls. Also moved the side effect of calling setHasCreatedUser directly after createUserProfile in the useEffect so that it is immediately set to true before anything else can happen.

- Granular useMemo for GraphQL Derived Data: Extracted the data transformations (platformRoles, accountPrivileges, accountEntitlements, accountId) into their own useMemo hooks. This isolates changes. If only the accountPrivileges change, only that useMemo will re-run, preventing unnecessary updates to providedValue.

- Simplified userMetadata useMemo: The userMetadata useMemo has been simplified to only depend on meData and platformLevelAuthorizationData and it calculates the myPrivileges and myRoles within the memoized function. This is more efficient.

- Precise providedValue Dependencies: The providedValue useMemo dependency array is now more targeted. It only includes the memoized values: userMetadata, accountId, platformRoles, accountPrivileges, accountEntitlements and the necessary flags that trigger loading or re-authentication, i.e., loading, loadingMeAndParentQueries, verified, isAuthenticated.

## How these changes reduce re-renders:

- By using granular useMemo, we prevent providedValue from being recalculated unless absolutely necessary. If only the accountPrivileges change due to a server update, only the accountPrivileges useMemo and providedValue useMemo will re-run.

- Caching strategy for the graphql data, so that the reference of meData, and platformLevelAuthorizationData will only change if the data changes.

## How these changes reduce HTTP requests:

- The hasCreatedUser state effectively prevents the createUserProfile mutation from being called multiple times for the same user. It ensures that the user creation logic is only executed once when the user first authenticates and their profile doesn't exist.

## Suggestions:

- Remove the **user creation** from the provider.
- Remove the fetch of entitlements, since they are used only in `useCreateSpaceLink` and in all of the 4 places where `useCreateSpaceLink` is imported entitlements are not being used.
- Extract the **account** data into a separate file (preferably a hook) and use it on requirement only since the data from `useAuthenticationContext` is used just once in places the user may not even go so this will exclude additional HTTP calls.

## Summary by CodeRabbit

- **New Features**
  - Introduced a state variable to track user profile creation status.
  
- **Refactor**
  - Enhanced user session management to ensure profiles are created only under the proper conditions.
  - Removed the `verified` property from user profile components for a simplified data structure.
  - Improved error handling and data retrieval logic for a more robust and reliable user experience.

### DIFF
--- a/src/domain/community/user/providers/UserProvider/UserProvider.tsx
+++ b/src/domain/community/user/providers/UserProvider/UserProvider.tsx
@@ -1,4 +1,4 @@
-import { FC, useEffect, useMemo, createContext } from 'react';
+import { FC, useEffect, useMemo, createContext, useState } from 'react';
 import {
   refetchUserProviderQuery,
   useCreateUserNewRegistrationMutation,
@@ -6,7 +6,7 @@ import {
   useUserProviderQuery,
 } from '@/core/apollo/generated/apollo-hooks';
 import { ErrorPage } from '@/core/pages/Errors/ErrorPage';
-import { AuthorizationPrivilege, LicenseEntitlementType, RoleName, User } from '@/core/apollo/generated/graphql-schema';
+import { AuthorizationPrivilege, LicenseEntitlementType, RoleName } from '@/core/apollo/generated/graphql-schema';
 import { useAuthenticationContext } from '@/core/auth/authentication/hooks/useAuthenticationContext';
 import { toUserMetadata, UserMetadata } from '../../hooks/useUserMetadataWrapper';
 
@@ -14,7 +14,7 @@ export interface UserContextValue {
   user: UserMetadata | undefined;
   accountId: string | undefined;
   loading: boolean;
-  loadingMe: boolean; // Loading Authentication and Profile data. Once it's false that's enough for showing the page header and avatar.
+  loadingMe: boolean;
   verified: boolean;
   isAuthenticated: boolean;
   platformRoles: RoleName[];
@@ -26,7 +26,7 @@ const UserContext = createContext<UserContextValue>({
   user: undefined,
   accountId: undefined,
   loading: true,
-  loadingMe: true,
+  loadingMe: true, // Loading Authentication and Profile data. Once it's false that's enough for showing the page header and avatar.
   verified: false,
   isAuthenticated: false,
   platformRoles: [],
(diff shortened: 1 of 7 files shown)


bobbykolev:
The changes are good. Although, I can't see big performance/timing gains these are nice optimizations! 👏

I can see the difference once we detach the account details. On local, the current query takes: 144ms, and without the account: 778ms - nice gains!

The `account` logic could be extracted as suggested.

Let's remove the redundant comments and we can converge this one.

src/domain/community/user/providers/UserProvider/UserProvider.tsx
```diff
@@ -51,51 +53,64 @@ const UserProvider: FC = ({ children }) => {
   });
 
   useEffect(() => {
-    if (isAuthenticated && !loadingMe && !meData?.me?.user) {
+    if (isAuthenticated && !loadingMe && !user && !hasCreatedUser) {
       createUserProfile();
```
bobbykolev, line 57: should we await here or initiation is sufficient?

reactoholic: No need of **await**, I have registered a new user successfully and logged with it.

src/main/topLevelPages/myDashboard/newVirtualContributorWizard/useVirtualContributorWizard.tsx
```diff
@@ -131,7 +131,7 @@ const useVirtualContributorWizard = (): useVirtualContributorWizardProvided => {
       allAccountSpaces: account?.spaces ?? [],
       availableSpaces: account?.spaces?.filter(hasCommunityPrivilege) ?? [],
     };
-  }, [data, user, targetAccount]);
+  }, [data, user, targetAccount]); // @bobbykolev, do we need `user` here a an array dependency?
```
bobbykolev, line 134: I"m not sure

bobbykolev (approved):

PR #7621, opened 2025-02-06
7611 - Memoization and optimization in the collaborative Excalidraw wrapper

reactoholic:
7611 - Memoization and optimization in the collaborative Excalidraw wrapper

### DIFF
--- a/src/domain/collaboration/whiteboard/containers/WhiteboardProvider.tsx
+++ b/src/domain/collaboration/whiteboard/containers/WhiteboardProvider.tsx
@@ -1,4 +1,4 @@
-import React, { PropsWithChildren } from 'react';
+import React, { useMemo, PropsWithChildren } from 'react';
 import { useWhiteboardFromCalloutQuery } from '@/core/apollo/generated/apollo-hooks';
 import {
   WhiteboardDetailsFragment,
@@ -36,12 +36,17 @@ const WhiteboardProvider = ({
     fetchPolicy: 'cache-and-network', // TODO: Check if this is still needed
   });
 
-  const callout = data?.lookup.callout;
+  const callout = useMemo(() => data?.lookup.callout, [data?.lookup.callout]);
 
-  const whiteboardContribution = callout?.contributions?.find(
-    contribution =>
-      contribution.whiteboard &&
-      (contribution.whiteboard.nameID === whiteboardId || contribution.whiteboard.id === whiteboardId)
+  // @@@ WIP ~ #7611
+  const whiteboardContribution = useMemo(
+    () =>
+      callout?.contributions?.find(
+        contribution =>
+          contribution.whiteboard &&
+          (contribution.whiteboard.nameID === whiteboardId || contribution.whiteboard.id === whiteboardId)
+      ),
+    []
   );
 
   const framingWhiteboard = callout?.framing.whiteboard;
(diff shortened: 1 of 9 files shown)


src/domain/common/whiteboard/excalidraw/CollaborativeExcalidrawWrapper.tsx
```diff
@@ -121,24 +131,85 @@ const CollaborativeExcalidrawWrapper = ({
...
+    window.addEventListener('scroll', handleScroll, { passive: true, capture: true });
 
     return () => {
-      handleScroll.cancel();
-      window.removeEventListener('scroll', handleScroll, true);
+      debouncedRefresh.current?.cancel();
+      window.removeEventListener('scroll', handleScroll, { capture: true });
     };
-  }, [handleScroll]);
+  }, [excalidrawApi]);
```
hero101, line 179: Depending on `excalidrawApi` will run `useEffect` twice.
Depending on `handleScroll` will run it once.

`excalidrawApi?.refresh();` will be transpiled to `(_a = excalidrawApi) === null || _a === void 0 ? void 0 : _a.refresh();` which is the same as using and if statement with neglectable performance benefit for the if statement, which you can do in the function definition also by the way.
Maybe profiling both approaches would shed some light.

IMO I wouldn't consider this change as a safer or faster option.

src/domain/collaboration/whiteboard/containers/WhiteboardProvider.tsx
```diff
@@ -36,12 +36,17 @@ const WhiteboardProvider = ({
...
-      (contribution.whiteboard.nameID === whiteboardId || contribution.whiteboard.id === whiteboardId)
+  // @@@ WIP ~ #7611
+  const whiteboardContribution = useMemo(
+    () =>
+      callout?.contributions?.find(
+        contribution =>
+          contribution.whiteboard &&
+          (contribution.whiteboard.nameID === whiteboardId || contribution.whiteboard.id === whiteboardId)
+      ),
+    []
```
hero101, line 49: dependency on `whiteboardId` just in case

src/domain/common/whiteboard/excalidraw/CollaborativeExcalidrawWrapper.tsx
```diff
@@ -121,24 +131,85 @@ const CollaborativeExcalidrawWrapper = ({
   const styles = useActorWhiteboardStyles();
 
   const { user } = useUserContext();
-  const username = user?.user.profile.displayName ?? 'User';
+  const username = useMemo(() => user?.user.profile.displayName ?? 'User', [user?.user.profile.displayName]);
```
hero101, line 134: I believe overdosing on `useMemo` can hurt more than it helps for inexpensive operations like this one.

hero101 (commented):
Some thoughts I have so far.

Note:
I am not sure but I don't believe passing memorized functions to children in the nested setup we have for the whiteboards would bring any benefit.
The passed on function can then wrapped by another, not memoized and passed further down that would reduce some re-renders but not cure much.

src/domain/common/whiteboard/excalidraw/CollaborativeExcalidrawWrapper.tsx
reactoholic: Yes, I forgot to revert this one, thanks for pointing it.

PR #7626, opened 2025-02-06
Excalidraw loadFiles behaviour improvements

hero101:
## The scene
I've done two pairs of benchmark - before and after of resource usage and a breakdown of running times.
The benchmark is pretty simple but very effective - one browser was just receiving, the other was sending events.
The receiving side was profiled for 20 seconds each time under heavy usage of the sender (a lot of events generated per second).

## Running times profiling
|before|after|
|---|---|
image|image

Looking at the `before` the reconciliation took `62.2%` (4486ms) of the total microtask time (11560ms).
The main issue I saw was that `loadFiles` took `99%` of the running time of the reconciliation (4452ms)
The profiling also highlighted the Excalidraw native `updateScene` takes half of the running time which is also a good candite for optimization. The only thing I can think of is calling it less but I suspect that will be hard to do - throttling the incoming events might be an improvement.

## The result
After a simple optimization of `loadFiles` I managed to take down it's running time down to 46ms - **97x!!!!** improvement; the total microtask time to 6803ms (41.2% improvement :arrow_upper_right:) and the reconciliation time **31x times!!!!** (144ms).

|---|before|after|improvement
|---|---|---|---
microtask|11560.4|6803.9|1.7x
reconcileElementsAndLoadFiles|4486.6|144.6|31x
loadFiles|4452.3|46.0|97x

---

## Resource profiling
|before|after|
|---|---|
image|image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@alkemio/client-web",
-  "version": "0.81.2",
+  "version": "0.81.2.excalidraw",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@alkemio/client-web",
-      "version": "0.81.2",
+      "version": "0.81.2.excalidraw",
       "license": "EUPL-1.2",
       "dependencies": {
         "@alkemio/excalidraw": "0.17.1-alkemio-8",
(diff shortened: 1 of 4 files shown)


bobbykolev (approved):

hero101:
drafting it before it's actually all hands tested

hero101:
closing in favor or https://github.com/alkem-io/client-web/pull/7635

PR #7641, opened 2025-02-07
Excalidraw throttling of events

hero101:
### The emission of the following events is throttled, meaning the client sends fewer events, thus less data
- mouse location events throttled to 10 times per second (30 previously)
- element mutation events throttled to 100 times per second (200 previously)

- client will not handle mouse location events until the scene is fully initialized

## todo
- fix version

## Summary by CodeRabbit

- **Chores**
  - Updated the client package version to a new release variant.
  - Adjusted timing for full scene and cursor updates to refine synchronization frequency.

- **Refactor**
  - Enhanced real-time collaboration on the whiteboard by shifting focus to scene initialization.
  - Implemented throttled scene synchronization for more efficient and stable updates.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@alkemio/client-web",
-  "version": "0.82.2",
+  "version": "0.82.2.ex-perf",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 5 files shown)


bobbykolev:
merged as part of release 10

PR #7653, opened 2025-02-10
Allow paste of images from internet in MD editor.

bobbykolev:
It looked straightforward, but @reactoholic please double-check it.

I think there was a redundant isImage check. We can probably further optimize and reduce code.

Note that I'm now able to paste entire block with text and images form the internet. 
However, when coping files locally, only one is uploaded.

### DIFF
--- a/src/core/ui/forms/MarkdownInput/MarkdownInput.tsx
+++ b/src/core/ui/forms/MarkdownInput/MarkdownInput.tsx
@@ -115,6 +115,10 @@ export const MarkdownInput = memo(
           return true; // Image
         }
 
+        if (clipboardData?.files?.length && clipboardData?.files[0].type.startsWith('image/')) {
+          return true;
+        }
+
         if (item.kind === 'string' && item.type === 'text/html') {
           const htmlContent = clipboardData?.getData('text/html');
           return htmlContent?.includes('<img') ?? false; // HTML tag with image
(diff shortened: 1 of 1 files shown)


reactoholic:
> It looked straightforward, but @reactoholic please double-check it.
> 
> I think there was a redundant isImage check. We can probably further optimize and reduce code.
> 
> Note that I'm now able to paste entire block with text and images form the internet. However, when coping files locally, only one is uploaded.

Thank you for your effort on this! The check you have removed in this PR is not redundant as it checks whether the pasted image is copied from the internet as HTML, but what you have wrote me here makes me wonder do we want such requirement - to copy text and image and paste it at once?

CC: @SimoneZaza - do we want our users to be able to copy text + images and paste them all together in our MD editor?

src/core/ui/forms/MarkdownInput/MarkdownInput.tsx
```diff
@@ -115,6 +115,10 @@ export const MarkdownInput = memo(
           return true; // Image
         }
 
+        if (clipboardData?.files?.length && clipboardData?.files[0].type.startsWith('image/')) {
```
bobbykolev, line 118: When I added this check it was more of a workaround. A know it works but I guess we can either take it out without iterating over all files/items or we should iterate over files separately. Otherwise, we get different items but if we have one file this method always returns true regardless of the item.

bobbykolev:
Closing as the implementation was transferred to: https://github.com/alkem-io/client-web/pull/7659

PR #7659, opened 2025-02-11
Client/7571 bugfix md editor paste

reactoholic:
## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced the handling of pasted images in the text input field, broadening detection criteria for images and improving upload reliability during paste operations.

- **Refactor**
  - Updated the layout of media insertion controls by replacing a fragment with a structured container for key buttons, resulting in a clearer and more consistent interface.

### DIFF
--- a/src/core/ui/forms/MarkdownInput/MarkdownInput.tsx
+++ b/src/core/ui/forms/MarkdownInput/MarkdownInput.tsx
@@ -115,6 +115,10 @@ export const MarkdownInput = memo(
           return true; // Image
         }
 
+        if (clipboardData?.files?.length && clipboardData?.files[0].type.startsWith('image/')) {
+          return true;
+        }
+
         if (item.kind === 'string' && item.type === 'text/html') {
           const htmlContent = clipboardData?.getData('text/html');
           return htmlContent?.includes('<img') ?? false; // HTML tag with image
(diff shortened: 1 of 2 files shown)


src/core/ui/forms/MarkdownInput/MarkdownInput.tsx
```diff
@@ -115,6 +115,10 @@ export const MarkdownInput = memo(
           return true; // Image
         }
 
+        if (clipboardData?.files?.length && clipboardData?.files[0].type.startsWith('image/')) {
```
bobbykolev, line 118: When I added this check it was more of a workaround. A know it works but I guess we can either take it out without iterating over all files/items or we should iterate over files separately. Otherwise, we get different items but if we have one file this method always returns true regardless of the item.

bobbykolev:
I like the simplifications. That way if we need to revise the functionality in the future, it will be quicker for understanding and applying changes. Let's improve the check around the files[0] which could be misleading and I think we're good to go!

src/core/ui/forms/MarkdownInput/MarkdownInput.tsx
reactoholic: Good point! I'll look at this.

bobbykolev (approved):
Approving without testing the latest changes. 
Let's see how it works on dev.

PR #7664, opened 2025-02-12
Whiteboard error handling

hero101:
The whiteboard will receive a message from the server that the save failed. This PR is handling the message.

If a message is "fail" received, the "last saved" timer won't change - showing the last actual save, and a small icon with a tooltip will appear showing that the save failed.
If a "success" message is received the timer will update and the icon will be hidden.

image

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -2510,7 +2510,7 @@
       "no-whiteboards": "No whiteboards have been created yet. Be the first to add one.",
       "locked-by": "{{user}} is currently editing this Whiteboard",
       "loadingScene": "Loading scene",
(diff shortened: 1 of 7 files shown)


hero101:
@SimoneZaza How does this look to you? Should we make it more noticeable or change thee approach?

hero101:
> @SimoneZaza How does this look to you? Should we make it more noticeable or change thee approach?

As agreed, I will leave only the piping for propagation of the error so we can build on top of it in the future.

hero101:
Travis is failing due to coveralls internal error, will bypass rules when approved.

reactoholic (changes requested):

reactoholic (approved):

PR #7670, opened 2025-02-13
[NameIds] Reduce the UrlResolver calls on every route change

ccanos:
Based on NameIds Branch. It should be Squashd!

Fixing the following: 
- [x] instead of applying the logic for every route (and use global filtering) use HoC only for the routes that needs URLResolver;
- [x] the loading of the UrlResolver was used in the user-facing components as it affects the views before fetching the data;
- [x] fix - subspace creation not working from DashboardNavigation;
- [x] fixes the product request to have only one Create button in the Dashboard Navigation;
- [x] fix the missing space icon in invites; 

The next one is not trivial, the fixes we've tried led to side effects and it is extracted as a separate item.
- fix the redundant calls for static tabs like space/community, or subspace/index, etc.

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced lazy-loading and error handling that deliver smoother page transitions.
  - Added a new tip in the guides section highlighting the platform’s Generative AI-powered support.

- **Refactor**
  - Improved loading state management across discussions, profiles, dashboards, and notifications for a more seamless experience.
  - Streamlined navigation and dashboard components, removing redundant elements for increased clarity and responsiveness.


ccanos (commented):
approved

PR #7771, opened 2025-02-28
6695 - Avatar support during vc creation

reactoholic:
## Summary by CodeRabbit

- **New Features**
	- Enhanced the Virtual Contributor profile creation process with updated on-screen text, now allowing users to choose an avatar immediately or personalize their profile later.
	- Introduced a new function for uploading avatars during the Virtual Contributor creation process.


src/main/topLevelPages/myDashboard/newVirtualContributorWizard/useVirtualContributorWizard.tsx
```diff
@@ -547,6 +559,18 @@ const useVirtualContributorWizard = (): useVirtualContributorWizardProvided => {
         accountId: myAccountId,
       });
 
+      if (avatar?.file && createdVc?.profile?.avatar?.id) {
```
bobbykolev, line 562: I didn't look into the initial implementation but from the changes in this PR, it looks like you can extract an uploadAvatar() function and call it in these 3 places. Or maybe run it just before the VC creation?

reactoholic: Yes, I can extract it, but I was trying to be fast. I will do it. Regarding running it before the VC creation - negative, we can't because the VC creation is returning me the bucket id which I am using to upload the visual to.

bobbykolev: Hm, I thought we were using temp storage. That was the idea of the temp storage - upload files before you have the entity/its StorageBucketId.

bobbykolev (approved):

Issue #7685, opened 2025-02-17
Use innovation flow to name the Space tabs


SimoneZaza:
## Description

As an admin, I want to decide the title of the different pages in my Space
As a user, I get confused when the page titles don't align with the content of the Space

## Acceptance criteria
Server
We will use the innovation flow on space level to store the data. 
The page tab titles become the phases and the content of this description block (blue block on the left side) the description of the phases.
- [ ] Migration to turn the current InnovationFlow of the Spaces that currently has just one phase called NOT_USED, into a structure usable for this:
   - Defaults to 4 states: "Dashboard", "Community", "Subspaces", "Knowledge Base", with the current descriptions of the top left blocks in the client of the 4 tabs.
   - Dashboard
   - Default description: Welcome to this Space! 
   - For Description field on the Dashboard of existing Spaces, **Copy** the content that is currently used (which used to be the context-Vision field, but is changed to be profile-description with #4944)
   - Community
   - Default description: This page displays all the members of this Space, including the host. If any guidelines have been set, you'll also find them here.
   - Subspaces
   - Default description: Here you can explore the hosted Subspaces. Filter by keywords to show just the topics you care about.
   - Knowledge Base
   - Default description: In the Knowledge Base you will find relevant documents, insights and other materials about the topic. 

- [ ] InnovationFlows have now constraints, minimum and maximum number of states. (TBD how to implement this serverside: IFs as JSON objects instead of just arrays of states?, new columns in the database?) 
- [ ] Users with UPDATE_INNOVATION_FLOW privilege in the innovation flow can change innovation flow (space admins should be able to)

Client
- [ ] Take the tab names from the innovation flow
- [ ] Do not try to load innovation flow if the user doesn't have permission - do not fail
- [ ] Default to the current names (Dashboard, Community, Subspaces and Knowledge Base) translated if they cannot be loaded for some reason
- [ ] Load the description of the state on the top left block.

This may complicate things and we may want to timebox it or separate it to another task, but in any case have it in mind during implementation:
- [ ] Innovation Flow's currentState will be the default tab selected when accessing a space's root url 

## Additional Context
Figma: https://www.figma.com/design/IOJdWGF5Bxmh69NUWl2Ws4/Info-Blocks-to-be-Editable?node-id=56-2&t=zpWuNPaArf0lvm6I-1

## Areas that will be affected 
 
_To be added during the refinement_

Issue #6937, opened 2024-09-24
New layout page in Space settings with Drag&drop editor


SimoneZaza:
## Description

As an admin, I want to decide what is stated in the description blocks on the different pages
As a user, I get confused when the descriptions don't align with the content of the Space

## Acceptance criteria
General
- [ ] The L0 (only Level 0) Space settings have a new tab: Layout
- [ ] The tab contents reuses the same component that we use for managing innovation flows. Drag and drop editor of callouts and states. + dialogs to change titles and descriptions.
- [ ] The First state will become the first tab, the second... and so on.
- [ ] The component to edit innovation flow now supports having a maximum and a minimum number of states and needs to handle (enable/disable) addition and deletion of states accordingly.
- [ ] On innovation flow changes, the tabs update immediately
- [ ] A property is added to the D&D editor's API so the term "phases" can be customized. (we want it says "Tabs" for spaces, "Phases" for subspaces. Confirm this with Simone at the moment of implementation

## Additional Context
Estimation only covering client work
Figma: https://www.figma.com/design/IOJdWGF5Bxmh69NUWl2Ws4/Info-Blocks-to-be-Editable?node-id=56-2&t=zpWuNPaArf0lvm6I-1

## Areas that will be affected 
 
_To be added during the refinement_

techsmyth:
@SimoneZaza suggest we split this one: allow at least the admins to have the edit in place, which is this issue. The client functionality is then tied to a privilege. And then there is a separate issue (epic) about redefining the roles in a community - which needs to be taken as a whole. Basically I am not keen to start giving the Leads additional privileges without considering the full set of roles...

SimoneZaza:
Yes, agreed!

bobbykolev:
Discussed and agreed on a refinement today:

- This implementation will be only for L0 spaces;
- The client will use the InnovationFlow states creating - HOME|COMMUNITY|SUBSPACES|KNOWLEDGE BASE - and will use their descriptions as requested;
- ⚠️ Important note - this mapping will be different than the one we have for the callout placement; 
- Using the state names as translated values in the Space tabs is nice to have (not planned but do it if easy);
- We'll have to create a migration creating these InnovationFlow states for all existing L0 spaces;

@SimoneZaza , 21 estimate because besides the _actual functionality of editing_, we also need to implement some _arch changes switching hardcoded fields to dynamic ones_ (using the states) but also because of the _migrations_ and potential risks around the callouts.

[...]

PR #7800, opened 2025-03-07
Feature: Flow states, about v2

techsmyth:
Only use flowState for the filtering of Callout. 

Removed all groups related functionality.

New hook to get the information needed for each of the tabs, including the state information that is then usable for both the tab name and the description.

Moved code around to have the tabbed pages as one layout under the new Space location.

Removed a couple of containers (merged into pages), and also a couple of hooks. 

Updated to use refined api on the server. 

Updated to use explicit passing around of classificationTagsets as an array

.....

Risk areas / TODOs:
- loading of content behind a dialog
- checking the routing for non-space pages
- getting the parameters for saving as template privilege (ideally from the library, so needs to be passed from even higher) and entitlement (needs collaboration license, also to come in from higher)
- ..
======

#7685
#6937
Disclaimer: There are certain hacks here to make the callouts be shown in the correct group when they come with certain flowState set that should NOT be merged. But serve as a proof of concept and display quite well the issue.

Problem:
- For positioning callouts in tabs we use `tagsets` that are in `callout > framing > profile > tagsets`.
- In spaces we use `tagset.name =  'callout-group'`. and there is a full server side implementation to filter and move callouts in between groups.
- In subspaces we use `tagset.name =  'flow-state'`. and we have a nice client-side implementation to move callouts between states with a drag&drop editor.
- Other server problems: Bad data and missing stuff, see the server PR https://github.com/alkem-io/server/pull/4995 for more context.

Second Problem:
- We have a very strict implementation for the tabs, we have 4 urls, dashboard, community, subspaces, knowledge base, that goes a bit deep, and it also uses a GraphQL type `CalloutGroupName`, we need to give flexibility to that, we can keep the 4/5 possible urls but I think we should unify DashboardPage, CommunityPage, SubspacesPage and KnowledgeBasePage and I think they should be a single one that renders the specific components for each section.

Server Actions: 
- I would implement some server side filtering at least, the same way we can query `callouts (groupNames[...]) {` to be able to query callouts by state.
- Groups code needs to be removed or adapted to use flow-states ?

Pending - things that may not be ver difficult and can be continued regardless of server changes
- Add description from the innovation-flow to top-right of the tabs
- Default to the `currentState` tab when we don't specify a tab in the url
- Add "Custom" tab (there is already some work on this. A fifth tab, so we'll have Dashboard (Home), Community, Subspaces, Knowledge-base and "Custom", when the page is activated through graphql.  There is some work already done on this. This custom page will be like knowledgeBase, just a list of callouts. Note that in the past there was a tab called Contribute, maybe some code can be reused.
- Remove the move callout dropdown from the calloutEditForm, we are not moving callouts like this any more.


ccanos:
@coderabbitai full review

PR #7802, opened 2025-03-07
Make the Join button consistent on the About Dialog L2

bobbykolev:
Fixing the issue in the ticket. Now the button on the about dialog should work similarly to the one behind on the dashboard.


Comoque1:
Tested OK locally.
Now the dialog is shown after clicking on "Join" button on about page.

PR #7810, opened 2025-03-11
7605 ~ Update search dialog

reactoholic:
### 7605 ~ Update search dialog

### DIFF
--- a/src/core/ui/card/cardsLayout/CardsLayout.tsx
+++ b/src/core/ui/card/cardsLayout/CardsLayout.tsx
@@ -17,6 +17,7 @@ export interface CardsLayoutProps<Item extends Identifiable | null | undefined>
   showCreateButton?: boolean;
   createButton?: ReactNode;
   globalSearch?: boolean;
+  loading?: boolean;
   onClickLoadMore?: () => void;
 }
 
@@ -30,6 +31,7 @@ export interface CardsLayoutProps<Item extends Identifiable | null | undefined>
  */
 const CardsLayout = <Item extends Identifiable | null | undefined>({
   items,
+  loading,
   children,
   deps = [],
   createButton,
@@ -67,7 +69,17 @@ const CardsLayout = <Item extends Identifiable | null | undefined>({
 
         {cards.length > 0 && (
           <Button
-            startIcon={<Autorenew />}
+            startIcon={
+              <Autorenew
+                sx={{
+                  animation: loading ? 'spin 1s linear infinite' : 'none',
+                  '@keyframes spin': {
+                    '0%': { transform: 'rotate(0deg)' },
+                    '100%': { transform: 'rotate(360deg)' },
+                  },
+                }}
+              />
+            }
             sx={{ minWidth: 'fit-content', height: 'fit-content' }}
             onClick={onClickLoadMore}
           >
--- a/src/core/ui/card/cardsLayout/CardsLayout.tsx
+++ b/src/core/ui/card/cardsLayout/CardsLayout.tsx
@@ -18,6 +18,7 @@ export interface CardsLayoutProps<Item extends Identifiable | null | undefined>
   createButton?: ReactNode;
   globalSearch?: boolean;
   loading?: boolean;
+  isButtonDisabled?: boolean;
   onClickLoadMore?: () => void;
 }
 
@@ -37,6 +38,7 @@ const CardsLayout = <Item extends Identifiable | null | undefined>({
   createButton,
   onClickLoadMore,
   globalSearch = false,
+  isButtonDisabled = false,
   ...layoutProps
 }: CardsLayoutProps<Item>) => {
   const { t } = useTranslation();
@@ -80,6 +82,7 @@ const CardsLayout = <Item extends Identifiable | null | undefined>({
                 }}
               />
             }
+            disabled={isButtonDisabled}
             sx={{ minWidth: 'fit-content', height: 'fit-content' }}
             onClick={onClickLoadMore}
           >
--- a/src/core/ui/card/cardsLayout/CardsLayout.tsx
+++ b/src/core/ui/card/cardsLayout/CardsLayout.tsx
@@ -1,8 +1,13 @@
+import { useTranslation } from 'react-i18next';
+
+import { Button } from '@mui/material';
+import { Autorenew } from '@mui/icons-material';
 import PageContentBlockGrid, { PageContentBlockGridProps } from '@/core/ui/content/PageContentBlockGrid';
 import { Identifiable } from '@/core/utils/Identifiable';
 import getDepsValueFromObject from '@/domain/shared/utils/getDepsValueFromObject';
 import { Box, BoxProps } from '@mui/material';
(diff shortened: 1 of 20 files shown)


src/root.tsx
```diff
@@ -32,6 +32,7 @@ import { UrlResolverProvider } from './main/routing/urlResolver/UrlResolverProvi
 // MARKDOWN_CLASS_NAME used in the styles below
 const globalStyles = (theme: Theme) => ({
   '*': {
+    scrollBehavior: 'smooth',
```
bobbykolev, line 35: Can we add this only where needed and not to all elements?

src/main/search/SearchView.tsx
```diff
@@ -68,230 +75,608 @@ interface SearchViewSections {
...
     return [];
   }
 
-  const spaceResults = (query.search.journeyResults || [])
+  const spaceResults = (query.search.spaceResults?.results || [])
     .map<SearchResultMetaType>(
-      ({ score, terms, ...rest }) => ({ ...rest, score: score || 0, terms: terms || [] } as SearchResultMetaType),
-      []
+      ({ id, type, score, terms, ...rest }) =>
+        ({ id, type, ...rest, score: score || 0, terms: terms || [] } as SearchResultMetaType)
```
hero101, line 604: Apart from the renaming, I am not convinced that changes are needed in this function.
- The additional `type` field is held by the `rest`.
- no sorting needed - the results are coming in sorted.

src/core/ui/card/cardsLayout/CardsLayout.tsx
```diff
@@ -40,6 +55,44 @@ const CardsLayout = <Item extends Identifiable | null | undefined>({
     [items, depsValueFromObjectDeps, children]
   );
 
+  if (globalSearch) {
```
hero101, line 58: I don't think it's ok to modify the layout in this way - It is a special single case.
Maybe have a component passed that does the "fetch more" functionality, or add a special card at the end of the array of items.
@bobbykolev Can I have a second opinion?

src/main/search/SearchView.tsx
```diff
@@ -68,230 +75,608 @@ interface SearchViewSections {
...
+          }
+
+          switch (resultsType) {
+            case SearchCategory.Spaces: {
+              return {
+                search: {
+                  ...fetchMoreResult.search,
+                  spaceResults: {
+                    ...fetchMoreResult.search.spaceResults,
+                    results: unionBy(
```
hero101, line 342: Is the `unionBy` necessary? There are no duplicating results.

src/main/search/SearchView.tsx
```diff
@@ -68,230 +75,608 @@ interface SearchViewSections {
...
+                      [
+                        ...(prev.search.spaceResults?.results ?? []),
+                        ...(fetchMoreResult.search.spaceResults?.results ?? []),
+                      ],
+                      'id'
+                    ),
+                  },
+                  calloutResults: prev.search.calloutResults,
+                  contributionResults: prev.search.contributionResults,
+                  contributorResults: prev.search.contributorResults,
```
hero101, line 352: is it required to build the whole object from scratch or can you just update the `results` field?

hero101 (changes requested):

src/core/ui/card/cardsLayout/CardsLayout.tsx
bobbykolev: I agree, it seems like you have everything you need (items.length, fetch etc) without the need of CardsLayout.  Something like:

```
<Gutters>
<CardsLayout ... />
<FetchMore ... />
</Gutters>
```

...without touching the CardsLayout, makes more sense. You can pass `sx` if needed.

___
Conceptually, the button could be perceived as part of the cards layout. Passing it as 'children' would also make sense. 

@reactoholic , feel free to elaborate on the challenges, pros and cons.

src/main/search/SearchView.tsx
reactoholic: I have added the `type` prop explicitly since TS was complaining, the rest of the logic is as it was implemented before, I haven't added anything else.

src/main/search/SearchView.tsx
reactoholic: It was necessary before since duplicated results were coming to the client.

src/main/search/SearchView.tsx
reactoholic: What do you mean by update just the "results" field, how? Everything else is using `prev`, but this is required in order the query to return the updated `results`.

src/main/search/SearchView.tsx
reactoholic: TypeScript is complaining when not having the `type` prop explicitly.

src/main/search/SearchView.tsx
hero101: I can suggest using immer, we have it in the project
```typescript
return produce(prev, next => {
  next.search.spaceResults = {
    // prefer incoming
    ...fetchMoreResult.search.spaceResults,
    results: [ ...prev.search.spaceResults.results, ...fetchMoreResult.search.spaceResults?.results ],
  }
});
```
or without it
```typescript
return {
  search: {
    ...prev.search,
    spaceResults: {
      // prefer incoming
      ...fetchMoreResult.search.spaceResults,
      results: [ ...prev.search.spaceResults.results, ...fetchMoreResult.search.spaceResults?.results ],
    }
  }
}
```

src/core/ui/card/cardsLayout/CardsLayout.tsx
reactoholic: > I agree, it seems like you have everything you need (items.length, fetch etc) without the need of CardsLayout. Something like:
> 
> ```
> <Gutters>
> <CardsLayout ... />
> <FetchMore ... />
> </Gutters>
> ```
> 
> ...without touching the CardsLayout, makes more sense. You can pass `sx` if needed.
> 
> Conceptually, the button could be perceived as part of the cards layout. Passing it as 'children' would also make sense.
> 
> @reactoholic , feel free to elaborate on the challenges, pros and cons.

I already spoke with Svetlio about that, you are not aware, but If I use the `Fetch more` button outside of `CardsLayout` the UI will broke in curtain situations depending on the user's viewport width. It's not not that ugly, but it's not quite cool also, thus the reason I moved this button inside of the layout. And If I try to write is per design with having the `fetch more` button outside `CardsLayout` will take more time.

src/main/search/SearchView.tsx
reactoholic: Oh, that's what you mean, okay! I will try it and check if it works without explicitly adding the rest of the object properties.

reactoholic: Just as I suspected - if we populate only the current section, the rest of the sections are deleted so we must provide the entire object of sections.

src/core/ui/card/cardsLayout/CardsLayout.tsx
hero101: @coderabbit Create an issue about refactoring the layout component. Use the details in the previous comments.

PR #7890, opened 2025-03-19
trying out another approach to route protection

techsmyth:
Not sure if this will fix the issue or not, but basic redirect if do not have read privilege is at least working...

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the way navigation handles user permissions for space sections.
  - Users without read access are now automatically directed to the overview page for a more consistent experience.


bobbykolev:
The approach doesn't solve the issue/s. It has pros and cons that need to be addressed. 
I'm closing as the protected route approach was taken and developed further in: https://github.com/alkem-io/client-web/pull/7899

PR #7918, opened 2025-03-21
return invitations when the inviting user is deleted

valentinyanakiev:
- quick solution
- the text may be in translation, but I am not sure yet whether this is overall the right approach (needs proper discussion)_
- goes with the same branch on the server
- overall making `createdBy` fieldResolver null & verifying the user exists

### DIFF
--- a/src/domain/community/pendingMembership/PendingMemberships.tsx
+++ b/src/domain/community/pendingMembership/PendingMemberships.tsx
@@ -97,14 +97,11 @@ export const InvitationHydrator = ({
   const userId = invitation.invitation.createdBy?.id;
 
   const { data: userData } = usePendingMembershipsUserQuery({
-    variables: { userId },
+    variables: { userId: userId! },
     skip: !userId,
   });
 
-  let userDisplayName: string = 'This user no longer exists on the platform!';
-  if (userData && userData.lookup.user?.profile.displayName) {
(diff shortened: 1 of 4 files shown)


src/domain/community/pendingMembership/PendingMemberships.tsx
```diff
@@ -120,7 +124,7 @@ export const InvitationHydrator = ({
         },
       },
     };
-  }, [invitation, journey, createdBy]);
+  }, [invitation, journey]);
```
hero101, line 127: add `userDisplayName` to deps

src/domain/community/pendingMembership/PendingMemberships.tsx
```diff
@@ -94,13 +94,17 @@ export const InvitationHydrator = ({
...
-    },
+    variables: { userId },
+    skip: !userId,
   });
 
-  const createdBy = userData?.lookup.user;
+  let userDisplayName: string = 'This user no longer exists on the platform!';
+  if (userData && userData.lookup.user?.profile.displayName) {
+    userDisplayName = userData.lookup.user?.profile.displayName;
+  }
```
hero101, line 107: `const userDisplayName = userData.lookup.user?.profile.displayName ?? 'This user no longer exists on the platform!';`

hero101 (changes requested):

hero101 (approved):

PR #7963, opened 2025-04-02
Refactoring: Space Admin, cards, structure tidy up

techsmyth:
NOTE: this PR touches so many files that it is essentially not reviewable. The bulk of the changes are either
* In the SpaceAdmin area: this is where code was actually refactored heavily
* Restructuring of the code base: moving things around, renaming

Key areas affected:
* SpaceAdmin area (fully tidied up), especially related to VC invitations
* Space Cards (light tidy up + renames)
* Deprecated items moved

___
TODO  (spotted):
PR and Spaces/Subspaces testing
- [x] fix the subspace cards (application not loading, space icons not present);
- [x] small pr fixes; 
- [x] Entitlements error on landing on a space without read access; 

Test invites:
- can't send another invite after there's a decline and the declined one is removed (not a PR issue);
- [x] invite VC to L2 space - all Acc VCs are available, adding resulting in error when not member of parent L1. 
___
## Refactorings

Refactored the full code around Space Admin.

Added typing on useCommunityAdmin hook

Removed deprecated hook for working with VCs

Passing + using about in multiple places to avoid the urlResolver duplicating effort

Removed "authorization" route on L1 that was not used. 

Refactored + simplified into a separate hook the VC interactions on community admin

Moved all deprecated items to be reviewed to a separate top level folder.

Tidied up the SpaceCards folder, including both removing one deprecated one and doing renamings / putting files into folders.

...

...


src/domain/space/layout/tabbedLayout/Tabs/SpaceDashboard/SpaceDashboardPage.tsx
```diff
@@ -33,7 +26,7 @@ const SpaceDashboardPage = ({ dialog }: PropsWithChildren<{ dialog?: TabbedLayou
       spaceId: spaceId!,
     },
     errorPolicy: 'all',
-    skip: !spaceId || spaceLevel !== SpaceLevel.L0 || !canReadSpace || loading,
+    skip: !spaceId || spaceLevel !== SpaceLevel.L0 || loading,
```
bobbykolev, line 29: This check prevented redundant requests when the spaceId changes to non-L0 and the space/tab layout is still in the DOM. Removing it might lead to fixing the same errors for the 'n'th time.

techsmyth: Indeed, but it is a very nasty smell that just has to be addressed....I can put it back in but it is just dealing with the symptom. There will continue to be subtle issues unless we fix the cause.

src/domain/shared/layout/EntityPageSection.ts
```diff
@@ -1,6 +1,6 @@
 export enum EntityPageSection {
   Home = 'home',
-  Dashboard = 'dashboard',
+  Dashboard = '',
```
bobbykolev, line 3: This was left intentionally, @ccanos can you remember why?

src/domain/space/context/SpaceContext.tsx
```diff
@@ -62,6 +69,15 @@ const SpaceContextProvider = ({ children }: PropsWithChildren) => {
     skip: !spaceId,
   });
 
+  const { data: spaceEntitlementsData, loading: loadingSpaceEntitlementsQuery } = useSpaceEntitlementsQuery({
```
bobbykolev, line 72: I'm not sure about awaiting the entitlements here. For the regular users, this could be a redundant wait.

src/domain/space/layout/tabbedLayout/Tabs/SpaceCommunityPage/SpaceCommunityPage.tsx
```diff
@@ -58,21 +57,17 @@ const SpaceCommunityPage = () => {
...
-    },
-    skip: !spaceId,
-  });
-
-  const spaceData = dataCommunityPage?.lookup.space;
-  const membership = spaceData?.about.membership;
+  const membership = about.membership;
   const communityId = membership?.communityID;
-  const communityGuidelinesId = dataCommunityPage?.lookup.space?.about.guidelines.id;
+  const communityGuidelinesId = about.guidelines!.id;
```
bobbykolev, line 62: I'm fixing this, but let's use less '!'. In the current situation, we might have a render before we have the id. So it's better to say '?.' and check where it is used, whether it exists.

src/domain/space/context/SpaceContext.tsx
bobbykolev: It could be other underlying issue but I hit an error with unauthenticated user trying to load space data:
image

bobbykolev: The error was fixed by skipping in case there's no read. But the question about the place for the entitlements remains.

PR #8036, opened 2025-04-14
Fix footer logo link

bobbykolev:
Target to R16 patch


bobbykolev:
change of plans, no patches as target

PR #8039, opened 2025-04-14
About Dialog polishing

bobbykolev:
- edit pencil and functionality in CGL;
- Apply button with helper text for unauthenticated;

Note that the CGL component has the edit icon in all places where used:
image


bobbykolev:
change of plans, no patches as target

PR #8047, opened 2025-04-14
Make sure the vc transparency block updates on every VC preview.

bobbykolev:
- now the VC profile is updating on every profile change preview, the ENGINE was missing from the VC data;
- the add to community was not returning a promise result, so the logic for updating the UI was moved into try-catch;
 

## Summary by CodeRabbit

- **New Features**
  - Expanded contributor profiles now include additional AI persona details, enriching the information available in community and admin views.

- **Refactor**
  - Streamlined the selection and invitation process for virtual contributors with improved state management and enhanced error handling for a more reliable user experience.


valeksiev (approved):
I think `InviteVcsDialog` needs a bit of refactoring. Two main points:
- it uses a ton of internal state which is not needed (I think) - some queries are executed -> components rerenders -> use effects are triggered that set some internal state -> component rerenders again
- event handlers don't follow the `handleEvent` naming convention

PR #8084, opened 2025-04-22
Add missing package for win

bobbykolev:
I'm having troubles running the clinent without installing this package. 
@techsmyth do you have similar issue running `npm run` after `npm i`?
@ccanos, can you double-check this is not causing issues on linux?


techsmyth:
@bobbykolev I found that I needed to delete package-lock and then re install - then it was ok...

PR #8088, opened 2025-04-22
Simplified code of ApplicationButton

ccanos:
The bug was not reproducible anymore, but I saw that it was fixed with the custom CSS: 
`sx={{ '&:hover': { color: theme => theme.palette.common.white } }}`

I have removed that and simplified the code. The Application button could be a `Buttton` or a `RouterLink`, but now they are always `Button`s, and that simplifies the ref, it doesn't have to be a Ref<Button | Link>.

Note for testing: The problematic button was the "Sign in to apply", so only visible when not logged in

## Summary by CodeRabbit

- **Refactor**
	- Simplified navigation for unauthenticated users by replacing link-based navigation with a standard button and click handler.
	- Streamlined ref handling by narrowing supported element types to button elements only.


hero101 (approved):
The button`s text is visible on hover.

PR #8090, opened 2025-04-22
Removes classification details from callout details query

valentinyanakiev:
The classification details are no longer needed in the
CalloutDetails query. This change removes the associated
fragments and logic to simplify the query and improve
performance by reducing the amount of data fetched. Also,
adds mutations for iframe allowed urls.


valentinyanakiev:
went for an alternative approach

PR #8107, opened 2025-04-24
Bulk mark as read notifications

bobbykolev:
Straightforward implementation -> icon, hook, mutation. I had to define a default state as the context was complaining.
The server code is already in develop branch.

## Summary by CodeRabbit

- **New Features**
  - Added a "Mark all as read" button to the in-app notifications dialog, allowing users to mark all unread notifications as read with a single click.

- **Bug Fixes**
  - Improved reliability of the notifications context to prevent issues when accessing notification state.

- **Localization**
  - Added English translation for the "Mark all as read" action in notifications.

### DIFF
--- a/src/main/inAppNotifications/useInAppNotifications.ts
+++ b/src/main/inAppNotifications/useInAppNotifications.ts
@@ -133,8 +133,10 @@ export const useInAppNotifications = () => {
         ID: id,
         state: status,
       },
-      update: cache => {
-        updateNotificationsCache(cache, [id], status);
+      update: (cache, data) => {
+        if (data?.data?.updateNotificationState === status) {
+          updateNotificationsCache(cache, [id], status);
+        }
       },
     });
   };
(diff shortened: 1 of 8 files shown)


src/main/inAppNotifications/useInAppNotifications.ts
```diff
@@ -111,5 +113,20 @@ export const useInAppNotifications = () => {
...
+
+    if (ids.length === 0) {
+      return;
+    }
+
+    await markAsRead({
+      variables: {
+        notificationIds: ids,
+      },
+      refetchQueries: [refetchInAppNotificationsQuery()],
```
hero101, line 127: refetching the whole set of notifications might be too heavy for just updating the state.

```suggestion
      update: cache => {
        const existingData = cache.readQuery<InAppNotificationsQuery>({
          query: InAppNotificationsDocument,
        });

        if (existingData) {
          const updatedNotifications = existingData.notifications.map(notification =>
            ids.includes(notification.id)
              ? { ...notification, state: InAppNotificationState.Read }
              : notification
          );

          cache.writeQuery({
            query: InAppNotificationsDocument,
            data: { notifications: updatedNotifications },
          });
        }
      }
```

hero101 (changes requested):

PR #8196, opened 2025-05-14
Esm module switch

techsmyth:
Node type switched to module (ES Module).

TSConfig updated to target ES2022 which is simply more modern (and has some benefits in terms of tree shaking, perf etc).

The linting + prettier integration is not yet working with the ESM setup; that was an area that was quite out of date...

To be finalized once the eslint + domain model updates tied to server api changes are merged.

## Summary by CodeRabbit

- **Chores**
	- Updated project configuration to use ES module syntax.
	- Added ES module designation in the project settings.
	- Upgraded TypeScript target to support ES2022 features.


techsmyth:
Created a new clean branch with what was in this that was different than eslint branch

PR #8227, opened 2025-05-20
Evict cache on roleSet modification

ccanos:


bobbykolev (dismissed):
Works as it should. 👏 I guess it makes something similar to fetchPolicy 'network cache or only', but it works like charm. Let's get rid of the commented code and merge.

PR #8229, opened 2025-05-20
Fix: host vs lead for orgs

ccanos:


bobbykolev (approved):
Approving, looking good. I'll update the ticket with one comment as we always show the host on the About dialog. It's a product requirement.

PR #8247, opened 2025-05-23
Organization Social Links

bobbykolev:
- [x] Fix Org references in settings (broken in develop);
- [ ] The Social Links block in Org Settings, similar to UserProfile Settings (requires server changes).

### DIFF
--- a/src/domain/common/profile/ProfileModelUtils.ts
+++ b/src/domain/common/profile/ProfileModelUtils.ts
@@ -11,7 +11,7 @@ type ProfileModelWithoutId = {
 export const mapProfileModelToUpdateProfileInput = (
   profileModel: ProfileModelWithoutId | undefined
 ): UpdateProfileInput => {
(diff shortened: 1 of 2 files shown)


src/domain/platform/admin/components/Organization/OrganizationPage.tsx
```diff
@@ -83,7 +83,7 @@ const OrganizationPage = ({ mode }: Props) => {
         },
       };
 
-      createOrganization({ variables: { input } });
+      await createOrganization({ variables: { input } });
```
ccanos, line 86: interesting that we had async but no await in here :S
I would return the promise, because there's no need to await for the result here

bobbykolev: resolving as you did

ccanos (approved):

PR #8256, opened 2025-05-28
Improves caching strategy for SPA

valentinyanakiev:
Enhances caching for Single Page Applications (SPAs) by implementing a more comprehensive no-cache strategy for `index.html` and other routes that serve it.
This change addresses issues with browsers and proxies caching outdated versions of the application.

The updated Nginx configuration includes rules for serving static assets with long-term caching, API routes without index.html fallback, and other routes (SPA routes) serving `index.html` with no-cache headers via a named location.

The Vite plugin is updated to aggressively prevent caching of `index.html` during development by setting more comprehensive no-cache headers. It now intercepts requests based on the route to apply the no-cache headers, excluding API and Vite internal routes.

### DIFF
--- a/.build/.nginx/nginx.conf
+++ b/.build/.nginx/nginx.conf
@@ -8,7 +8,7 @@ http {
         root  /usr/share/nginx/html;
         include /etc/nginx/mime.types;
 
-        # Prevent caching of index.html
+        # Prevent caching of index.html when accessed directly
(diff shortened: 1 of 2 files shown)


bobbykolev (approved):
Looking good.
We should remember where to add exclusions for files if we want to cache them. 

@ccanos FYI

PR #8257, opened 2025-05-28
Change switch in Subspaces form to add or skip callouts of the template

ccanos:
There's no Subspace tutorials in the server
There's only space tutorials:
image

So this switch makes no sense:
image

I have changed it to:
image

## Summary by CodeRabbit

- **New Features**
  - Updated subspace creation form to use a new "Add default template's collaboration tools" option, replacing the previous tutorial callouts field.
  - The new option is enabled by default and features a dynamically translated label.

- **Bug Fixes**
  - Ensured consistent handling of the new collaboration tools option throughout the subspace creation workflow.

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -862,6 +862,9 @@
         "placeholder": "",
         "tooltip": "Tags"
       },
+      "addCallouts": {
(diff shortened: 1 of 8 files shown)


bobbykolev (approved):
looking good, not tested *

PR #8260, opened 2025-05-29
app version detection logic init

bobbykolev:
Documentation: 
https://github.com/alkem-io/client-web/blob/004d83c8b45b04ec4f0ba1d7e54d6737e2fbd1a6/docs/service-worker-version-handling.md

- Create meta.json on build (holding the latest package.json version). It shouldn't be cached!
- Register a service worker.
- Pass the app version from the client to the SW.
- The worker checks the latest built version available under /meta.json (public/assets).
- If there's a version mismatch, a message is sent to the client. 
- The client currently shows a notification toast. 
- Design and functionality of the notification.

## Summary by CodeRabbit

- **New Features**
  - Introduced automatic detection of new app versions using a service worker and version metadata.
  - Added user notifications prompting a page refresh when a new version is available, including a reload button.
  - Implemented a React component for handling version update notifications.
  - Generated a version metadata file during build to support version checks.

- **Documentation**
  - Added detailed documentation explaining the service worker version handling system.

- **Chores**
  - Updated build configuration to generate version metadata.
  - Updated ignore rules to exclude version metadata from version control.

- **Refactor**
  - Simplified and improved service worker registration and version communication logic.
  - Replaced direct Snackbar usage with a custom notification component for better flexibility.


public/service-worker.js
```diff
@@ -0,0 +1,90 @@
...
+  if (!currentVersion) {
+    return;
+  }
+
+  try {
+    const response = await fetch(VERSION_URL, { cache: 'no-store' });
+    const data = await response.json();
+    const newVersion = data.version;
+
+    if (newVersion !== currentVersion && isMajorOrMinorChanged(currentVersion, newVersion)) {
```
valeksiev, line 38: Are we assuming we don't need to refresh on patch updates? Do we need to know the type of version bump or just check if the versions are different?

valeksiev (dismissed):

PR #8345, opened 2025-06-13
hard reload after version mismatch to bypass Safari's cache

bobbykolev:
## Summary by CodeRabbit

- **Refactor**
	- Improved the page refresh mechanism by updating the reload button to use a more efficient navigation method, resulting in a smoother user experience when refreshing the page.


valeksiev (approved):
Hard to test locally but the code simulates a redirect so should be fine.

PR #8349, opened 2025-06-16
[WIP] Callouts refresh

ccanos:


ccanos:
@coderabbitai full review

src/core/utils/nameOf.ts
```diff
@@ -1,19 +1,12 @@
-// const nameOf = <T>(name: Extract<keyof T, string>): string => name;
```
ccanos, line 1: previously this nameOf function had another purpose, it was only used for the urlParams, I have moved this code there, and implemented a new `nameOf` util, which is more similar to the c# nameOf util, it's useful for forms, see the comment in code

src/core/ui/content/PageContentBlockCollapsible.tsx
```diff
@@ -1,26 +1,58 @@
...
 import { ReactNode, forwardRef, useState } from 'react';
 import PageContentBlock, { PageContentBlockProps } from './PageContentBlock';
 import { gutters } from '../grid/utils';
+import PageContentBlockSeamless from './PageContentBlockSeamless';
+
+interface ExpandMoreButtonProps extends IconButtonProps {
+  collapsed: boolean;
+}
+
+const ExpandMoreButton = styled((props: ExpandMoreButtonProps) => {
```
bobbykolev, line 12: we can extract this as a separate reusable component

PR #8356, opened 2025-06-17
WiP: Updates in L1 & L2 settings

bobbykolev:
- [x] L2 admin settings updated to be like L1 (excluding Create Subspace);
- [ ] ... 
- [ ] ...

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -2156,7 +2156,7 @@
             "createBlocks": "<b>Create Blocks</b>: Allow members to create blocks for collaboration tools & posts. If you deactivate this, members will still be able to contribute to existing posts and calls, but will not be able to create new blocks.",
             "createSubspaces": "<b>Create Subspaces</b>: Allow members to create Subspaces in this Space. If you deactivate this, you can still create Subspaces yourself and add other people as an admin inside the Subspace.",
             "eventsFromSubspaces": "<b>Subspace events</b>: Allow events from Subspaces to be visible on this Space calendar as well.",
-            "inheritRights": "Allow Space members to contribute.",
+            "inheritRights": "Allow members of to contribute without needing to join first.",
             "supportAsAdmin": "<b>Alkemio Support</b>: Allow the Alkemio Support team to act as an admin in this Space."
           },
           "copySpace": {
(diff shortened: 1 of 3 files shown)


src/domain/community/inviteContributors/components/FormikContributorsSelectorField/useContributors.ts
```diff
@@ -0,0 +1,103 @@
...
+  } = usePaginatedQuery<UserSelectorQuery, UserSelectorQueryVariables>({
+    useQuery: useUserSelectorQuery,
+    getPageInfo: data => data.usersPaginated.pageInfo,
+    options: { skip: !filter || onlyUsersInRole },
+    pageSize,
+    variables: { filter },
+  });
+
+  if (onlyUsersInRole && parentSpaceId) {
+    const data = useMemo(
```
ccanos, line 85: I don't think you can useMemo inside this if

src/domain/community/inviteContributors/components/FormikContributorsSelectorField/useContributors.ts
```diff
@@ -0,0 +1,103 @@
...
+  pageSize = 20,
+}: UseContributorsParams): UseContributorsResult => {
+  const { data: parentSpace, loading: loadingRoleSet } = useInviteUsersDialogQuery({
+    variables: {
+      spaceId: parentSpaceId!,
+    },
+    skip: !onlyUsersInRole || !parentSpaceId,
+  });
+
+  const parentRoleSetId = parentSpace?.lookup.space?.about.membership.roleSetID;
```
ccanos, line 61: not sure how hard would be the changes, but it would be very nice to pass the parentRoleSetId directly instead of the `parentSpaceId`
Also `onlyUsersInRole` would be a complex structure: { roleSetId, roleName, contributorType } and use it if it's set or use the userSelector if it's null

ccanos (changes requested):
please see the comments

src/domain/community/inviteContributors/components/FormikContributorsSelectorField/useContributors.ts
bobbykolev: As we have the parentId provided, it's better to make the additional query when needed. No need to obtain the roleSetId earlier. You have a point and if you insist on passing the roleSetId instead, then whenever we set the `onlyUsersInRole`, we'll have to make sure we're querying and passing the roleSetId instead.

PR #8365, opened 2025-06-19
Fix the import of a space templates.

bobbykolev:
Dependant on: https://github.com/alkem-io/server/pull/5200


src/domain/templates/components/TemplatesAdmin/TemplatesAdmin.tsx
```diff
@@ -264,7 +268,7 @@ const TemplatesAdmin = ({
         ...template,
         spaceId: template.contentSpace?.id,
       });
-      await createSpaceTemplate({
+      await importSpaceTemplate({
```
techsmyth, line 271: is there a new Template entity being created? If so then it is create...

PR #8426, opened 2025-07-02
Optimized global banner

bobbykolev:
The Global banner is the Largest Contentful Paint (LCP) item on the homepage.
In this PR, the asset was optimized, reducing its size from ~250kb to ~190kb without visible changes.
https://optimize.svgomg.net/

This small change would potentially improve our LCP score/timing, improving the perf score in Google (not only).

image


ccanos (approved):

ccanos:
incredible, I can't find a single pixel difference... nice work 👏

PR #8482, opened 2025-07-09
SubSpace Creation - populate the form with tmpl data

bobbykolev:
- on Tmpl selection the form is populated with the template data;
- on template removal, only the templateId and the visuals are removed, the other data can be removed manually by the user;
- in case the form is 'dirty', there's a confirmation dialog;
  - note that the override confirm text for posts was altered to work with any Form;

Note that the FormikVisualUpload component was extended to support the display of assets by providing a URL. 

## Summary by CodeRabbit

* **New Features**
  * Enhanced template import to prefill form fields and visuals (avatar, card banner) when creating a subspace.
  * Added a confirmation dialog to prevent accidental overwriting of unsaved changes during template import.

* **Improvements**
  * Visual upload fields now support displaying initial images from selected templates.
  * Template selection now loads additional profile details, including tagline, tags, and visuals.
  * Updated confirmation dialog text for template import for clearer messaging.

### DIFF
--- a/src/core/i18n/en/translation.en.json
+++ b/src/core/i18n/en/translation.en.json
@@ -1571,6 +1571,7 @@
         "title": "Confirm {{entity}} Deletion",
         "description": "Are you certain you want to delete this {{entity}}? Keep in mind that this action is irreversible. <strong>You can only delete a {{entity}} if there are no underlying Subspaces associated with it.</strong> If there are any Subspaces within this page, please delete them first (you can do so from the {{entity}} settings > Subspaces page).",
         "descriptionShort": "Are you certain you want to delete this {{entity}}? Keep in mind that this action is irreversible.",
+        "descriptionShortWithName": "Are you certain you want to delete <strong>{{entity}}</strong>? Keep in mind that this action is irreversible.",
         "checkbox": "Please check this box if you are certain about deleting this {{entity}}",
         "confirm": "Yes, delete"
       }
--- a/src/domain/shared/components/EntityConfirmDeleteDialog.tsx
+++ b/src/domain/shared/components/EntityConfirmDeleteDialog.tsx
@@ -11,6 +11,7 @@ import { Trans, TransProps, useTranslation } from 'react-i18next';
 
 interface EntityConfirmDeleteDialogProps {
   entity: string;
+  name?: string;
   description?: TranslationKey;
   open: boolean;
   onClose: () => void;
@@ -19,6 +20,7 @@ interface EntityConfirmDeleteDialogProps {
 
 const EntityConfirmDeleteDialog: FC<EntityConfirmDeleteDialogProps> = ({
   entity,
+  name,
   description,
   open,
   onClose,
@@ -50,7 +52,7 @@ const EntityConfirmDeleteDialog: FC<EntityConfirmDeleteDialogProps> = ({
                   (description ??
                     'components.deleteEntity.confirmDialog.description') as unknown as TransProps<TranslationKey>['i18nKey']
                 }
-                values={{ entity: entity }}
+                values={{ entity: name ?? entity }}
                 components={{ strong: <strong /> }}
               />
             </Caption>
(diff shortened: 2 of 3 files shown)


src/domain/templates/components/TemplateSelectors/SubspaceTemplateSelector.tsx
```diff
@@ -46,13 +77,60 @@ export const SubspaceTemplateSelector: FC<SubspaceTemplateSelectorProps> = ({ na
...
+      }
+
+      setDialogOpen(false);
+    }
   };
 
   const handleRemoveTemplate = (): void => {
+    // reset the template field and the visuals
     helpers.setValue('');
+    onTemplateVisualsLoaded?.(getVisualUrls([]));
```
ccanos, line 120: this is the same as `onTemplateVisualsLoaded?.({})`
you may need `{} as BasicVisualUrlModel` but I think it's worthit to save the call to get VisualUrls

src/domain/templates/components/TemplateSelectors/SubspaceTemplateSelector.tsx
```diff
@@ -1,26 +1,55 @@
...
+};
 
 interface SubspaceTemplateSelectorProps extends GuttersProps {
   name: string;
+  onTemplateVisualsLoaded?: (visualUrls: BasicVisualUrlModel) => void;
 }
 
-export const SubspaceTemplateSelector: FC<SubspaceTemplateSelectorProps> = ({ name, ...rest }) => {
+const getVisualUrls = (visuals: VisualModel[]): BasicVisualUrlModel => {
+  const avatar: string | undefined = getVisualByType(VisualType.Avatar, visuals)?.uri;
```
ccanos, line 33: It's not exactly the same result because your way is not even setting the property if it's `undefined`, but it doesn't really matter if it's undefined, I think something like this would be simpler and easier to read: 
```
return {
   avatar: getVisualByType(...),
   cardBanner: getVisualByType(...),
};
```

src/domain/templates/components/TemplateSelectors/SubspaceTemplateSelector.tsx
```diff
@@ -1,26 +1,55 @@
...
+  useTemplateContentLazyQuery,
+} from '@/core/apollo/generated/apollo-hooks';
 import { Identifiable } from '@/core/utils/Identifiable';
 import Gutters, { GuttersProps } from '@/core/ui/grid/Gutters';
 import useUrlResolver from '@/main/routing/urlResolver/useUrlResolver';
+import { getVisualByType } from '@/domain/common/visual/utils/visuals.utils';
+import { VisualModel } from '@/domain/common/visual/model/VisualModel';
+import ConfirmationDialog from '@/core/ui/dialogs/ConfirmationDialog';
+
+export type BasicVisualUrlModel = {
```
ccanos, line 22: I think we could make this a common thing. I'd move it do `domain/common/visual/utils`
And call it something like `EntityVisualUrls`
have there the four types:
```
export type EntityVisualUrls = {
  avatarUrl?: string;
  cardBannerUrl?: string;
  bannerUrl?: string;
  bannerWideUrl?: string;
}
```
and the function `getVisualUrls` moved together there

src/domain/space/components/subspaces/CreateSubspaceForm.tsx
```diff
@@ -139,26 +151,21 @@ export const CreateSubspaceForm = ({
           />
           <Gutters padding={theme => `${gutters()(theme)} 0 0 0`}>
             <PageContentBlock sx={{ flexDirection: isSmallScreen ? 'column' : 'row', justifyContent: 'space-between' }}>
-              <FormikVisualUpload name="visuals.avatar" visualType={VisualType.Avatar} flex={1} />
-              <FormikVisualUpload name="visuals.cardBanner" visualType={VisualType.Card} flex={1} />
+              <FormikVisualUpload
+                name="visuals.avatar"
+                visualType={VisualType.Avatar}
+                flex={1}
+                initialVisualUrl={templateVisuals.avatar}
```
ccanos, line 158: I don't see the reason to have a `showInitialVisual`, is there any reason to not have just the string, and if the string is undefined or empty interpret it as if `showInitialVisual` was `false`?

ccanos (changes requested):
a few small comments, happy to discuss about them

PR #8493, opened 2025-07-11
New Space Visuals 

bobbykolev:
Initially, I misunderstood the requirements. Instead of having the same default asset for all subspaces based on the top-level space, we should just have a random* but persistent one. The logic should depend only on the space's ID no matter the SpaceLevel and parent-child relations. 

So we should have 16 variations (0-f based-id), the logic for displaying is based on the first character in the space ID.
There's one additional default, which will be displayed in case there's no ID provided - on creation or during loading.

The main logic is in `src/domain/space/icons/defaultVisualUrls.ts`. 

## Summary by CodeRabbit

* **New Features**
  * Default images for spaces (avatars, banners, cards) are now dynamically generated based on each space’s unique ID, providing more variety and context-aware visuals.
  * Space IDs are now consistently included in space-related components, improving identification across cards, avatars, tiles, banners, and notifications.

* **Bug Fixes**
  * Improved fallback logic for default images, ensuring the correct image is shown when a custom visual is unavailable.

* **Refactor**
  * Updated multiple components to use a new function for determining default visual URLs, replacing previous static mappings.
  * Enhanced prop interfaces for space-related components to support the new dynamic image logic.


ccanos:
@bobbykolev I have optimized a bit the getDefaultVisual function, I think when executed many times, regexes can be expensive and a Set like this is faster.

PR #8495, opened 2025-07-11
updated codegen for new mutations for working with innovation flow states

techsmyth:
Graphql mutations added for working with states on innovation flow

To do:
- update code to use the new mutation
- test...


techsmyth:
@ccanos am hoping this is 80% plus, but good if we have a short chat early next week so I can talk you through the updates. Basically the states are now managed much more like callouts in a calloutsSet, so you can add, remove + re-order - all different mutations.

PR #8536, opened 2025-07-28
Multiselect for the invite dialog

bobbykolev:
Now the Invite Dialog supports multiple selection of roles. 

A new reusable FormikMultiSelect control was introduced.

Side note, I'm planning to extract all of the helper text logic into a util, but I didn't want to be part of this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a multi-select dropdown component for forms, allowing users to select multiple options with improved UI elements.

* **Enhancements**
  * Updated the invite users dialog and form to support selecting multiple roles instead of just one when inviting contributors.
  * Improved validation and handling for multiple role selections during user invitations.
  * Refined invitation logic to consistently handle multiple roles and exclude default member role automatically.

* **Bug Fixes**
  * Ensured consistent handling of role selection and invitation logic throughout the user interface.

### DIFF
--- a/src/core/ui/forms/FormikMultiSelect.tsx
+++ b/src/core/ui/forms/FormikMultiSelect.tsx
@@ -1,0 +1,89 @@
+import { Autocomplete, Chip, TextField, TextFieldProps } from '@mui/material';
+import { useField } from 'formik';
+import { useMemo, ReactElement, SyntheticEvent } from 'react';
+import { useValidationMessageTranslation } from '@/domain/shared/i18n/ValidationMessageTranslation';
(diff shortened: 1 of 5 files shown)


ccanos:
@bobbykolev Code looks good and I like the UI, but I think it's confusing to be able to remove "Member" from the select because it's not true, the person will become a member anyway.
image
what if we add a flag "mandatory" or "notRemovable", or just "disabled" to the options, so that option cannot be removed?
I have seen this in the docs:
image

ccanos (changes requested):
please see the comment

ccanos (approved):

Issue #8545, opened 2025-07-30
BUG: Navigating from Welcome page to Sign In page, throws console errors


Comoque1:
**Describe the bug**

**To Reproduce**

1. Go to welcome.dev-alkem.io
2. Click on "Sign in" button - user is redirected to sign in page and console error is thrown
`KratosInput.tsx:57 Warning: Failed prop type: The prop `xs` of `GridLegacy` can only be used together with the `item` prop.`
<img width="1305" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd3pqQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--98e706d0091a6396710d47ff34032e77f35afb35/image.png" alt="image.png" />

**Expected behavior**
Console error is not thrown.

PR #8621, opened 2025-08-21
Fix: Navigating from Welcome page to Sign In page, throws console errors #8545

ccanos:
I was not seeing the console error anymore but it makes sense, in the docs when you set any breakpoint to a Grid you are referring to a GridItem.
I started adding `item` as prop to GridLegacy and saw no difference visually.
So I have removed <GridLegacy> completely. Tested it visually and looks exactly the same. KratosUI (the parent that renders all the `KratosInput`s is not an MUI Grid anyway.

Tested a few screens: Sign In, Sign Up, set password, Enter email verification code, reset password (maybe not all) but tested forms look exactly the same.
So one less GridLegacy item to migrate when we upgrade MUI 🎉

<hr/>
Also fixes this other console error:
image

## Summary by CodeRabbit

* **Refactor**
  * Simplified the input layout: inputs now render as a TextField wrapped by a Tooltip (grid wrapper removed). Visual spacing/alignment and responsive behavior may be slightly different; typing, focus/blur, validation, error display, helper text, disabled state, and password visibility remain unchanged.
  * Improved component rendering to ensure stable React key handling with no user-visible behavior changes.

### DIFF
--- a/src/core/auth/authentication/components/Kratos/KratosInput.tsx
+++ b/src/core/auth/authentication/components/Kratos/KratosInput.tsx
@@ -1,4 +1,4 @@
-import { GridLegacy, InputAdornment, OutlinedInputProps, TextField, Tooltip } from '@mui/material';
+import { InputAdornment, OutlinedInputProps, TextField, Tooltip } from '@mui/material';
 import { UiNodeInputAttributes } from '@ory/kratos-client';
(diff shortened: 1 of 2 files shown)


bobbykolev (approved):
Nice, please double-check mobile as well.

ccanos:
sure, did it. Merging, thanks

PR #8672, opened 2025-09-04
Improved read-only reason for Memo

hero101:
## Summary by CodeRabbit

- **New Features**
  - Memo editing now respects content update policies, restricting edits when only contributors may modify content.
  - Read-only reasons are clearer and include policy-based restrictions.
  - Read-only state is detected earlier for more consistent behavior with connectivity/auth checks.
  - Members with appropriate access can continue editing as before.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@alkemio/client-web",
-  "version": "0.109.3",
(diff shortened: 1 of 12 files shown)


bobbykolev (approved):
Just double-check the conditional order. Otherwise, looking good.

PR #8736, opened 2025-09-17
Ability to disable user geo location

hero101:
## Summary by CodeRabbit

- New Features
  - Added a geolocation enable/disable toggle in configuration.

- Bug Fixes
  - App now skips location lookup and clears loading/data when geolocation is disabled or no endpoint is configured.
  - Analytics always use numeric coordinates (fallback to 0) to avoid undefined values.

- Refactor
  - Internal logic now reacts to geolocation enable/disable changes without altering public interfaces.


src/core/analytics/geo/UserGeoProvider.tsx
```diff
@@ -29,13 +29,17 @@ export const UserGeoContext = createContext<UserGeoContextProps>({
 export const UserGeoProvider = ({ children }: PropsWithChildren) => {
   const { geo } = useConfig();
   const geoEndpoint = geo?.endpoint;
+  const disabled = !geo?.enabled;
```
valeksiev, line 32: Same as the server - wouldn't invert the flag.

valeksiev (approved):

PR #8748, opened 2025-09-18
Tiptap: Upgrade to 3+

polibb:

Upgrades tiptap's packages to latest version.
- using `clientId` to distinguish users suddenly does not work as expected - 1 user has many different clientIds for all their contributions in a memo, so changed to use user's `id` itself
- removed some configuration flags which we override by using explicit extensions (link extension; history from extension-collaboration is called undo/redo and included)
- all tiptap packages are now 3.4.4

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added


ccanos:
I see a problem, not sure if it's here only, 
image

when collaborating, the cursor looks like this. I remember we copied the styles from the tiptap website to `src/core/ui/forms/CollaborativeMarkdownInput/styles.scss` they may have changed the css classes or something

in dev it looks like this:
image

ccanos (changes requested):
there's a styling problem, other than that, looks to be working well! good job.
But there's a change about userIds... in the end we wanted the server to add the userId to the information broadcasted and not to be the client who informs the server saying "this is my user id". I think it can stay like you have done and I'll create a task to refactor a bit this, Svetlio and me already had it in mind

polibb:
Thank you for spotting the styling error! Fixed ☕️

> But there's a change about userIds... in the end we wanted the server to add the userId to the information broadcasted and not to be the client who informs the server saying "this is my user id". I think it can stay like you have done and I'll create a task to refactor a bit this, Svetlio and me already had it in mind

I was wondering why the TODO was there instead of just using the userId.. Now I get it. After the update of tiptap the field of `id` does not accept numbers, which the `clientId` is, and also there are many clientIds per user depending on their session collaboration input. So yeah, it has to be userId now, and a refactoring task is a good idea, thank you.

PR #8881, opened 2025-10-17
isContactable fixes

bobbykolev:
## Summary by CodeRabbit

* **New Features**
  * Added contactability status for contributors and users, enabling control over contact visibility in profiles and contributor cards throughout the application.

* **Version Update**
  * Released version 0.116.0

### DIFF
--- a/src/domain/community/user/userCard/UserCard.tsx
+++ b/src/domain/community/user/userCard/UserCard.tsx
@@ -42,7 +42,7 @@ const UserCard = ({
   url,
   roleName,
   loading,
-  isContactable = true,
+  isContactable = false,
   onContact,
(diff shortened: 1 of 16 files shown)


src/domain/community/user/userCard/UserCard.tsx
```diff
@@ -42,7 +42,7 @@ const UserCard = ({
   url,
   roleName,
   loading,
-  isContactable = true,
+  isContactable = false,
```
ccanos: we may miss some envelop icons in places where they should be, 
but at least we don't show anymore envelop icons where they shouldn't be

ccanos (approved):

Issue #8146, opened 2025-05-04
BUG: p under p not allowed


techsmyth:
**Describe the bug**
Displaying activity results in the following error: 
<img width="801" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL1RsQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a283ca517f84609380af6d737240780dd09b6718/image.png" alt="image.png" />

**To Reproduce**

1. Create a new space
2. Go to the space and click on the button to show all activity
3. Toggle the switch to show "most active" 
4. Above error is thrown in the consoe

**Expected behavior**
No error is thrown in the console 

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

**Areas that will be affected**
_Optional, to be added by developer if relevant_

PR #8946, opened 2025-10-28
A few console errors fixed

ccanos:
Couldn't reproduce #8146

But I have fixed a few other console errors I've found trying to reproduce it:
image

image

## Summary by CodeRabbit

* **Updates**
  * Refined tooltip behavior in dashboard navigation when interacting with space avatars and settings.


polibb (approved):
tested, looks good

PR #9407, opened 2026-03-11
simple robots.txt file, remove redundant build logic, update the env.sh

bobbykolev:
Note the target: [021-add-robots-txt](https://github.com/alkem-io/client-web/tree/021-add-robots-txt)

Removes ~88 lines of build logic from buildConfiguration.js, simplifies tests, and aligns with the project's existing runtime env var injection pattern.

Having the robots.txt file in plain text is better. Easier to read, understand and update (from product as well).

The VITE_APP_ROBOTS_ALLOW_INDEXING renamed (omit _APP), no need to 'leak' into the DOM.

⚠️ Requires ENV variable setup in infra ops.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -13,13 +13,13 @@ Allow: /
 
 # Sensitive / authenticated-only areas
 Disallow: /admin
-Disallow: /identity
 Disallow: /restricted
-Disallow: /profile
+Disallow: /user/me
 
 # API and internal endpoints
 Disallow: /api/
 Disallow: /graphql
+Disallow: /graphiql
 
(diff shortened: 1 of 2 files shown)


public/robots.txt
```diff
@@ -0,0 +1,96 @@
...
+# Global rules (all well-behaved crawlers)
+# ----------------------------------------------
+User-agent: *
+Allow: /
+
+# Sensitive / authenticated-only areas
+Disallow: /admin
+Disallow: /identity
+Disallow: /restricted
+Disallow: /profile
```
valentinyanakiev: don't think this exists, it is /user/xxx

PR #9421, opened 2026-03-16
Simplified url resolver

bobbykolev:
Context

 The current UrlResolverProvider sends the full URL to the server, which parses it, resolves all nameIds to UUIDs, checks permissions at every level, and returns a flat bag of ~20 IDs. This creates three problems:

 1. Blocking rendering - Pages show a <Loading /> spinner until the entire URL is resolved (then again while space data loads)
 2. Server owns route knowledge - The server duplicates client route patterns via path-to-regexp matchers in url.resolver.service.ts
 (1100+ lines). Route changes require coordinating both codebases.
 3. Monolithic coupling - A single context (UrlResolverContextValue) bags 20+ fields; every consumer depends on it regardless of which entity they need.

 Goal: Client-owned routes, non-blocking page rendering, clean separation of business logic from views (preparing for UI library switch).

- [x] spec for review
- [ ] implementation


techsmyth:
@bobbykolev lets chat this one through next week. As shared in the chat, I do not agree that the client should own the routes as there are multiple ways in which the URLs are used / generated.  We may want to carefully define a URL scheme for *this particular* client. It is an open question if the client needs to call to the server for URL resolution - if that was dropped would that help?

Issue #9512, opened 2026-04-08
[CRD] Search modal redesign


bobbykolev:
## Description

As a **platform user**,
I want the Search dialog redesigned using the CRD design system (shadcn/ui + Tailwind),
So that platform search is visually consistent with the new UI, feels faster and more polished, and follows the improved UX patterns from the prototype (Cmd+K trigger, animated overlay, scroll-tracked category sidebar, cleaner result cards).

The current search dialog (`src/main/search/SearchDialog.tsx`) uses MUI's `DialogWithGrid`, `DialogContent`, and MUI-based result cards (SpaceCard, ContributingUserCard, etc.) with a chip-based multi-term input (`MultipleSelect`).
It opens via URL query params (`?search-terms=...`) and displays results in 4 categories: Spaces & Subspaces, Collaboration Tools, Contributions, Contributors.

  The prototype provides a complete redesign reference at `prototype/src/app/components/search/SearchOverlay.tsx` — a full-viewport animated overlay with 5 categories (Spaces, Posts, Responses, Users, Organizations), a category
  sidebar with scroll tracking and result counts, per-section filter dropdowns, scope switching (platform vs current space), and dedicated result cards per type. Triggered via Cmd+K / Ctrl+K or header search icon.

  ## Acceptance criteria

  - [ ] Toggle CRD on — Global Administration → Layout (`localStorage.setItem('alkemio-crd-enabled', 'true')`)
  - [ ] Search opens as a full-viewport animated overlay (not MUI dialog), with backdrop blur and scale-in animation
  - [ ] Cmd+K / Ctrl+K keyboard shortcut opens search (in addition to existing header trigger)
  - [ ] Escape key closes search
  - [ ] Search input with tag-based multi-term entry (Enter to add, X to remove tags, max 5 terms)
  - [ ] Scope dropdown visible when inside a space — toggles between "All Spaces" and current space
  - [ ] Results display in categorized sections: Spaces, Posts, Responses, Users, Organizations
  - [ ] Category sidebar (desktop) with scroll-tracking active state, result counts per category, click-to-scroll
  - [ ] Category pills (mobile) — horizontal scrollable tabs replacing sidebar on small screens
  - [ ] Per-section filter dropdowns: Spaces (All / Spaces only / Subspaces only), Posts (All / Whiteboards / Memos), Responses (All / Posts / Whiteboards / Memos)
[...]

PR #9588, opened 2026-04-27
[CRD] Search in Space/Platform

### LINKED ISSUE #1844: [CRD] Search in Space or Platform
## Description

  As a user browsing inside a Space
  I want a scope selector in the CRD search overlay that lets me choose between searching the
  current Space or the entire platform
  So that I can find Space-specific results without leaking platform-wide noise — matching the
   MUI search behavior that the CRD migration in #9518 missed

## Acceptance criteria

  - Should detect the current (level-zero) Space via the app's Space context (not URL regex),
  so scoping works on /<spaceNameId> and nested Subspace routes
  - Should show a "Search In:  | Entire platform" dropdown in the search bar when on any Space
   tree page, and hide it on non-Space pages
  - Should default the active scope to the current Space when the overlay opens on a Space
  page, and restrict results to that Space tree
  - Should re-execute the query automatically (with the same tags) when the user switches
  scope, showing the loading state
  - Should never apply a space filter when the active scope is "Entire platform"
  - Should show a "Search the entire platform instead" recovery button in the no-results panel
   only when scoped to a Space, and one click flips scope + re-runs the query
  - Should reset the scope to the page default every time the overlay closes (no persistence
  across opens)
  - Should not fire the query or show a stale/placeholder Space label while the Space
  id/displayName are still resolving
  - Should rename the existing search.scopeAll value to "Entire platform" across all six
  languages (en, nl, es, bg, de, fr) and add new crd-search keys for scope/recovery copy in
  all six
  - Should keep the scope selector keyboard-operable and screen-reader-announced (focus, arrow
   keys, Escape, accessible name reflecting active scope)
  - Should keep all CRD architectural rules: presentational dropdown stays in src/crd/ with no
   MUI/Apollo/domain/router imports; detection + state lives in src/main/crdPages/search/
  - Should leave the MUI search dialog untouched when the CRD toggle is off

## Additional Context

  - Spec: specs/093-crd-search-scope/spec.md
  - Root cause of current bug: extractSpaceNameIdFromPath in CrdSearchOverlay.tsx matches
  /space/<id> but real Space URLs are /<spaceNameId> — so on-Space searches today silently
  return platform-wide results
  - Context source note: CrdSearchOverlay is mounted above SpaceContextProvider, so useSpace()
   is not available there — integration must use useUrlResolver + useSpaceAboutBaseQuery (see
  plan.md / research.md)
  - Subspaces always scope to the level-zero parent Space (matches MUI)
  - Visual dropdown already exists: SearchTagInput accepts scope / onScopeChange — just needs
  wiring through SearchOverlay → CrdSearchOverlay

polibb:
Follow-up to #9518 (issue #9512). The CRD search overlay was missing the scope-switching behavior the MUI version has, and it had a latent bug that made the scoping silently broken even where it was meant to work.

## Two gaps closed:

  1. Bug fix — Space detection never matched real URLs. The integration tried to detect the current Space using a regex against /space/<id>, but Alkemio Space URLs are /<spaceNameId>
  directly. As a result, on every Space page the search ran platform-wide instead of being scoped — verifiable in the original issue: searching wb on /secondspace returned posts from both
  Welcome Space and second space. The regex is removed; the integration now reads levelZeroSpaceId from useUrlResolver() (the same source SpaceContextProvider itself uses) and looks up the
  Space's display name with useSpaceAboutBaseQuery — works on top-level Spaces and Subspaces alike.
  2. Feature — scope dropdown in the search bar. When the user is browsing inside any Space tree, the overlay now shows a "Search In: …" pill with two options: the current top-level Space's
  display name (default) and "Entire platform". Switching re-fires the query without the user re-typing. When the user is not on a Space page, no dropdown is rendered and search is
  platform-wide as before.

  2. Plus a "Search the entire platform instead" recovery action in the no-results panel for one-click widening when a scoped search returns nothing.

  2. Behavior mirrors MUI 1:1 (matching wording, "Search In:" prefix, dark-navy pill).

## Additional context

  - Spec / plan / tasks in specs/093-crd-search-scope/ (spec, plan, research, data-model, contracts, quickstart, tasks, analyze report).
  - Subspace pages show the level-zero Space, not the Subspace, matching MUI's useSpace()-driven model. This was clarified in the spec via /speckit.clarify.
  - i18n: the existing search.scopeAll value renamed from "All Spaces" → "Entire platform" in all six languages to match MUI's components.search.scope.platform. New search.scopeTriggerLabel
  ("Search In: {{option}}") and search.a11y.scopeTrigger keys added in en, nl, es, bg, de, fr.
  - No GraphQL schema changes, no codegen, no new dependencies. The existing searchInSpaceFilter variable is reused; only its value changes (level-zero Space UUID when scoped, undefined when
   platform-wide).
  - Z-index fix: the dropdown menu's Radix portal renders to document.body with z-50 by default — that's below the overlay's z-[100]/z-[101] backdrop, so the menu was opening behind the
  backdrop and silently swallowing clicks. Override to z-[110] at the call site.
  - MUI fallback path is untouched: toggling CRD off still renders the legacy search dialog with no behavior change (SC-007).

## Governance

  - Documentation is added — full spec + plan + research + data-model + contracts + quickstart under specs/093-crd-search-scope/
  - Test cases are added or updated — new CrdSearchOverlay.test.tsx (6 Vitest + RTL tests covering FR-007 / FR-012 / FR-003-004 / FR-016 / FR-017 / FR-019, plus a regression guard for the
  on-/home "search never fires" bug)

## By submitting this pull request I confirm that:
  - I understand that any contributions or suggestions I made may make it into the actual code. I've read the License https://github.com/alkem-io/Coordination/blob/master/LICENSE.
  - I understand that submitting the pull request requires agreeing to and signing the contributor license agreement

## New Features
* Added scope selector to search overlay enabling users to choose between searching the current space or the entire platform, with automatic re-execution on scope changes.
* Improved search accessibility with proper aria-labels for scope selection and appropriate keyboard navigation support.
* Scope selection resets when closing and reopening the search overlay.

## Internationalization
* Updated search UI terminology across six languages to clarify scope options ("Entire platform" vs space-specific search).


polibb:
@coderabbitai review

PR #9611, opened 2026-04-30
[CRD] Check if user is beta tester to show the switch or not

ccanos:


bobbykolev:
dropped as requirement

PR #9662, opened 2026-05-13
[CRD] P0: Space fixes

### LINKED ISSUE #9660: [CRD] P0: Space fixes
## Description

- Whiteboard contributions open the collaborative editor directly with no inline preview step,
  unlike posts which show a read-only preview first.
- The contributions section and existing contributions disappear when both Members/Admins
  contribution switches are turned off, instead of just hiding the Add tile.
- The post edit dialog has no UI for adding or removing references on existing posts.
- Users with `MovePost` privilege have no way to move a post contribution to another
  post-collecting callout within the same Space/Subspace.
- Switching framing or response type to "None" in edit mode silently clears the existing
  memo/document/whiteboard/post-collection on save, with no warning that the action is irreversible
   and locks out other types.
- The callout form's tags field is a comma-separated text input instead of a chip-based tag
  input.
- Removing a reference row in the references editor happens immediately with no confirmation.
- Closing the callout detail dialog with unsaved comment text loses the text silently.
- Post cards don't surface the callout's external references or default tags.
- On contribution post cards, hidden tags behind the "+N more" pill are unreachable — there's no
  way to see them without opening the post.
- The framing editor doesn't show an inline thumbnail of the configured whiteboard.
- The post-preview timestamp shows an absolute date instead of a relative "x ago" distance,
  duplicating the header timestamp.
- The memo dialog uses two different inline loading treatments instead of the shared `Loading`
  component.
- Can save a post callout with a tag of 1 character

## Acceptance criteria

- [ ] Should be fixed - work as expected from MUI

## Additional Context
Screen shots / UX designs / data that are relevant.

polibb:

 - **Callout detail dialog — whiteboard inline preview**: clicking a whiteboard contribution now
  shows an inline preview (header + thumbnail + click-to-open overlay) before launching the
  collaborative editor, matching MUI behavior. Closing the editor returns to the preview, not the
  grid.
  - **Contributions section visibility**: decoupled the contributions section/grid visibility from
  `settings.contribution.enabled`. Existing contributions stay visible when contributions are
  soft-disabled; only the Add tile is gated on `canCreateContribution` (MUI parity).
  - **Post contribution dialog — references editing**: post edit dialog now supports
  adding/removing references on existing posts (diffed against server state; uses
  `createReferenceOnProfile` + `deleteReference`).
  - **Post contribution dialog — post location**: added the "Post location" dropdown in edit mode
  for users with `MovePost` privilege, allowing moving a post to another post-collecting callout
  within the same callouts-set.
  - **Callout form — confirmation when clearing framing/response type**: switching framing or
  response type to "None" in edit mode now triggers a confirmation dialog, since the action
  permanently removes the existing memo/document/whiteboard/post-collection and the lock prevents
  adopting a different type afterwards.
  - **Callout form — tags use `TagsInput`**: replaced the comma-separated text input with the
  chip-based `TagsInput` component.
  - **References editor — remove confirmation**: removing a reference row now requires
  confirmation.
  - **Callout dialog — close confirmation**: closing with unsaved comment text triggers a
  confirmation dialog.
  - **Card — references + tags strip**: new `ReferencesAndTagsStrip` component rendered on
  `PostCard` (and reused in the detail dialog), showing external references as links and
  default-tagset tags as pills.
  - **Contribution post card — "+N more" tag tooltip**: hidden tags are now reachable via a hover
  tooltip on the `+N` pill (MUI parity).
  - **Framing editor — inline whiteboard preview**: added inline thumbnail preview of the
  configured whiteboard using the last-saved preview blobs.
  - **Date formatting**: post-preview timestamp now uses relative ("1 hour ago") formatting via a
  new `formatRelativeFromNow` helper, matching MUI's contribution-preview header.
  - **Memo dialog — loading polish**: unified the two loading/connecting states to use the shared
  `Loading` component.
- Cannot save a post callout with a tag of 1 character
  - **i18n**: added keys for the new flows (`postLocation.*`, `whiteboardPreview.*`,
  `clearTypeConfirm.*`, `confirmClose.*`, `references.removeConfirm.*`, `contributionsHeader`)
  across all 6 supported languages (en, nl, es, bg, de, fr).

### Additional context

Add any other context

### Governance 

- [ ] Documentation is added

## Release Notes

* **New Features**
  * References and tags now display on posts and callout details
  * Added whiteboard contribution previews with inline editing
  * Enhanced tags input to support multi-tag entry and minimum length validation
  * Introduced post location/movement capability to reassign posts between callouts
  * Added relative time formatting for recent contributions

* **Improvements**
  * References removal now requires confirmation to prevent accidental deletion
  * Closing dialogs with unsaved changes triggers a confirmation prompt
  * Whiteboard framing type is now locked during edits to maintain consistency
  * Tag display updated with collapsible overflow indicators and tooltips
  * Enhanced editor responsiveness with improved click target sizing

* **Internationalization**
  * Added translations for references, tags, post movement, and confirmation dialogs across all supported languages

[Review Change Stack](https://app.coderabbit.ai/change-stack/alkem-io/client-web/pull/9662)


polibb:
@coderabbitai

 Applied (still-valid issues)

  File: src/crd/components/callout/InlineWhiteboardPreview.tsx
  Fix: Discriminated-union props so deleteLabel is required when onDelete is set; trash button's aria-label now uses deleteLabel directly.
  ────────────────────────────────────────
  File: src/crd/components/callout/ReferencesAndTagsStrip.tsx
  Fix: Added role="list" (with biome-ignore matching the project's existing pattern) to restore list semantics after Tailwind reset; +N pill is now <Badge asChild><button type="button">… instead of a div with tabIndex={0}.
  ────────────────────────────────────────
  File: src/crd/i18n/space/space.nl.json
  Fix: confirmClose.description: "reactietekst" → "opmerkingstekst" (matches the file's other comments.* strings, all of which use "opmerking").
  ────────────────────────────────────────
  File: src/main/crdPages/post/CrdPostContributionDialog.tsx
  Fix: (a) Modified id-bearing references are detected via id+content diff and converted to delete+create instead of being silently dropped. (b) Move detection now uses a mutable baselineCalloutIdRef, updated on successful move so subsequent saves don't re-fire the
    move.
  ────────────────────────────────────────
  File: src/main/crdPages/space/dataMappers/contributionDataMapper.ts
  Fix: toDateString and mapAnyContributionToCardData accept an optional Locale.
  ────────────────────────────────────────
  File: src/main/crdPages/space/callout/CalloutDetailDialogConnector.tsx
  Fix: (a) Three catch-all else branches replaced with explicit === Whiteboard checks so Link contributions don't get routed to the whiteboard editor. (b) Reset effect on dialog close clears all per-contribution state and overlay flags. (c) Threads i18n.language →
    date locale into mapAnyContributionToCardData.
  ────────────────────────────────────────
  File: src/main/crdPages/space/callout/ContributionsPreviewConnector.tsx
  Fix: Threads i18n.language → date locale into mapAnyContributionToCardData.
  ────────────────────────────────────────
  File: src/main/crdPages/space/callout/ResponseDefaultsConnector.tsx
  Fix: useEffect clears previewImages and resets whiteboardPreviewSettings when the dialog closes or type switches away from 'whiteboard'.

  Skipped (with reasons)

  - ContributionPostCard.tsx +N badge → <button> — The outer card is already a <button type="button">. Wrapping the inner +N Badge in a <button> would create nested interactive elements (invalid HTML / a11y tree breakage). The clean fix would require restructuring
  the card itself out of being a button — out of scope. Removing only the tabIndex={0} would silently lose keyboard discoverability of the hidden tags. Left as-is intentionally; the same pattern in ReferencesAndTagsStrip (which is not nested in a button) was fixed.
  - CalloutFormConnector.tsx "clear to none" payload — Verified MUI parity in src/domain/collaboration/callout/CalloutDialogs/EditCalloutDialog.tsx:282: MUI explicitly strips allowedTypes from the update payload with the comment "AllowedTypes is read-only for now".
  The server doesn't accept response-type changes after creation, so emitting an explicit clear from the mapper would not be honored. The current omission matches MUI parity. The dubious piece is the form's "switch to None" UI flow itself, which is a separate
  UX/backend concern.

  Self-review pass

  I re-read each touched file looking for additional issues:
  - CalloutDetailDialogConnector reset-on-close correctness: the if (open) return; guard makes the effect idempotent across re-renders, and the dependency array correctly latches the latest initial* props as the reset values when the dialog actually closes.
  - ResponseDefaultsConnector reset effect: confirmed it preserves state during the editor's save cycle (open=true, type='whiteboard' → early return) and only clears on close or response-type change.
  - CrdPostContributionDialog move flow: after a failed move the baseline isn't updated, so a subsequent save retries — desired. After a successful move, baseline updates, and the next render's isMoveTargetChanged becomes false.
  - InlineWhiteboardPreview: confirmed three <button>s are DOM siblings inside a <div> (not nested), so the discriminated union is the only soundness fix needed.
  - formatShortDate callers: only one call site (contributionDataMapper), now locale-aware.

PR #9691, opened 2026-05-18
Design version in tracking

hero101:


bobbykolev (approved):
Nice, clean solution.
Not tested but looking good.

If it goes with the patch we might have a bit slower adoption (until users refresh their browser session). The last one is just for awareness, not related to the changes.

PR #9725, opened 2026-05-22
Fix Markdown lists mismatches

ccanos:
## The problem

Some markdown content was rendering weird: a single bulleted list showed up as **two separate `<ul>`s**, the items in one of them came wrapped in `<p>` tags while the others didn't. Looked broken and inconsistent. The markdown content had a mix of `*` and `-`:

```md
* aaaaa1 ...

- bbbbb1 ...

- bbbbb2 ...
```

was rendered to

```html
<ul>
  <li>aaaaa1</li>
</ul>
<ul><!-- note these ones have p's -->
  <li><p>bbbbb1</p></li>
  <li><p>bbbbb2</p></li>
</ul>
```

## Concepts I've learnt today:
- "Loose" list item and "tight", In HTML and in Word documents etc, you can have multiple paragraphs inside a List Item, that's a loose list item.
  In markdown you can't because a new line in the list item directly takes you to another element. The markdown renderer marks List items as loose if they have a new empty line after (they end with `\n\n`). And list items that end with just `\n` and then comes directly another `-` (list item) are tight.
  The markdown renderer, puts a `<p>` inside loose list items (the ones that have a new line after), but in lists that only have one single item they are always tight.
```md
- this is a tight list item



No matter how many new lines come after, this will render as <ul><li>this is a tight list item</li></ul>
```

## The reason behind:

I'd say the rendering is behaving correctly, the problem is on the markdown we have saved, which is already inconsistent. Two things combine:

1. **Mixed bullet markers (`*` then `-`)** - CommonMark treats a marker change as a brand new list, so you get two `<ul>`s instead of one.
2. **Blank lines between items** - that makes a list "loose", and loose lists wrap every item in a `<p>`. A list with a single item can't be loose, so it stays "tight" (no `<p>`). That's exactly why one list had `<p>`s and the other didn't.

So who wrote that bad markdown? **Our own editor.** When you save, we go tiptap HTML → markdown via remark-stringify, and:

- remark-stringify **deliberately** alternates the bullet marker between two lists that come together (`*` then `-`) so they don't merge into a single one on. That's where the mixed bullets come from.
- tiptap wraps every `<li>` in a `<p>`, which makes remark write blank lines between items → loose lists.

## How to reproduce
Probably happens when pasting from Word or an email or website, but an easy way to reproduce is:
- Create a bullet list of 1 element
- Press enter once (a second bullet appears), press enter again (the list is over, the cursor is back to the left)
- Type a text
- Create another bullet list of multiple elements
- Remove the text in between. 

Now you have multiple `<ul>'s` that are going to be converted to markdown alternating `*` and `-` to separate them
And the lists who only have 1 element in them are going to be rendered as (tight) `<ul><li>(content)</li></ul>` with no `<p>` in them.

image
(Notice the different font of `aaaa1`)

## What I changed and why

Fixed it at the **save path** (`UnifiedConverter`, the HTML→Markdown step). Added a small remark plugin that runs right before we serialize to markdown and:

- **merges directly-adjacent lists of the same type into one** => no more marker alternation and the two-`<ul>`s split.
- **forces all lists tight** => no more `<p>` wrapping inside `<li>`.
- **changed the style of the `li`s** => Because before all the tight `li`'s where wrongly styled.

#### Why fix the save path and not the render side?
That style change is already gonna make it look better but it's still going to show a bit of separation between different `<ul>`s

#### Why tight and not loose: 
Because markdown doesn't support multiple paragraphs inside list-items anyway. Tight is the normal, compact form, and we already control the spacing between items via CSS - so going loose would just double up the spacing and risk `<p>` styling leaking into list items.

Lists separated by any content (with paragraph between them) still stay separate - only directly-adjacent ones merge.

## What to keep an eye on

- **Existing content stays "bad" until it's reopened and saved.** It's not auto-migrated - that's on purpose, but worth knowing. a big and quite risky migration would be required, but without migrating it's going to look better already.
- **Multi-paragraph list items** The editor can produce them, (Shift + Enter), that has worried me a bit cause maybe we need loose list-items, but I've been testing it and the markdown is cleaner now, it's adding `\n` to the list items.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved list formatting consistency when converting between HTML and Markdown formats.
  * Adjacent lists of the same type are now properly merged into a single list.
  * Lists now render in tight formatting by default for better readability and consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/alkem-io/client-web/pull/9725?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


techsmyth (approved):
Works great for me, tested on the same content as had the issue and now displaying with consistent fonts :)

Note: did not need to open + save it, displayed straight away with consistent fonts.

ccanos:
> Note: did not need to open + save it, displayed straight away with consistent fonts.

It does display better existing texts yeah, that's the change in the css (`root.tsx`). We could go just with that but,
before, the texts directly inside a `<li>` were getting the wrong font so a markup like:
```
<ul>
<li>Content 1</li>
</ul>
<ul>
<li><p>Content 2a</p></li>
<li><p>Content 2b</p></li>
</ul>
```
was displaying wrong. From now it will style the first `<li>` well, but it will leave a small space between `<ul>`s.
The list normalizer will join any lists that have nothing in between them.

PR #9840, opened 2026-06-10
translations against msg house + UK english

techsmyth:
Ran the EN + Nl translations against the message house 

NL suggested updates are in a md file, which needs then to be removed before merging.

### DIFF
--- a/nl-crowdin-updates.md
+++ b/nl-crowdin-updates.md
@@ -1,0 +1,47 @@
+# Dutch (NL) Crowdin updates — message-house alignment
+
+These changes must be made in **Crowdin** (the NL `translation.nl.json` is Crowdin-managed and
+must not be edited directly in the repo). Driver: canonical glossary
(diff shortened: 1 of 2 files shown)


bobbykolev:
Hi @techsmyth , in CRD we are now using only agents and translated modules. You can review and apply changes directly with specs here. 
Crowdin is used only in the OLD UI. We are not supporting it at the moment. 

Pass the `client-web/docs/crd/migration-guide.md` to the agent/speckit for details on the translations.

SimoneZaza (approved):

techsmyth:
@bobbykolev Simone is ok with the translation changes, not a high priority but if you can take this one through from here that would be great...

bobbykolev:
Hi both, I'm closing this one in favour of that one: https://github.com/alkem-io/client-web/pull/9859
It's not a partial but full overhaul of all translations. + some cleanup of Crowdin.

PR #9986, opened 2026-07-07
feat(017-combined-subspace-application): offer subspace Apply via ROLESET_ENTRY_ROLE_APPLY + setting copy

valentinyanakiev:
## Summary

**UI slice of `workspace#017-combined-subspace-application`** (alkem-io/server#6189): a non-parent-member viewing a reachable Subspace is offered **"Apply"** directly — the client trusts the server-granted `ROLESET_ENTRY_ROLE_APPLY` privilege instead of inferring eligibility from `isParentMember`.

### Changes

- `useApplicationButton`: `canApplyToCommunity` / `canJoinCommunity` now derive purely from the `RolesetEntryRoleApply` / `RolesetEntryRoleJoin` privilege on the subspace role-set (dropped `&& isParentMember`). The parent-first prompts (`PreApplicationDialog` / `PreJoinParentDialog`) remain the fallback **only when the privilege is absent** — per the contract in workspace ADR 0001 §1.
- New unit test: privilege-driven gating independent of `isParentMember`; fallback preserved without the privilege (4/4 green).
- `spaceSettings.{en,nl,es,bg,de,fr}.json`: the `subspaceAdminInvitations` setting description now states it governs **applications as well as invitations** (FR-010/SC-006, key parity across all 6 locales).

**No GraphQL change** — the privileges were already selected via `membership.myPrivileges`; no codegen needed.

### Rollout

Depends on the server slice (alkem-io/server#6232) exposing the privilege — server ships **first**. A non-updated server simply never grants APPLY to non-parent-members, so this change is inert until then; no broken intermediate state.

### Verification

- `pnpm build` clean, `pnpm lint` clean, unit test 4/4.
- Live-verified end-to-end against the local stack running the server slice (Apply CTA for a non-member, questions render, application appears in the Subspace community, updated setting copy visible).

Refs alkem-io/server#6189 · `workspace#017-combined-subspace-application` · companion: alkem-io/server#6232

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the logic for showing join/apply options so they better match your permissions and community setup.
  * Clarified invitation and application messaging in multiple languages, including when a user can join after accepting an invitation or having a request approved.


valentinyanakiev:
## Adversarial code review (`/review`)

Three angles over the full diff (line-by-line, removed-behavior tracing across all consumers, i18n parity) with self-refutation. **No findings survived.**

What was verified:
- **Boolean reduction is exact**: old `canApply = priv && (!isSubspace || isParentMember)` → new `priv` changes exactly one case — `isSubspace && !isParentMember && priv` (was `false`, now `true`) — which is precisely the intended relaxation (server only grants the privilege when the combined preconditions hold).
- **All consumers traced** (`SpaceAboutApplyButton` → `useSpaceApplyFlow` cascade): the parent-first prompts remain reachable whenever the server withholds the privilege (private subspace / setting off); the L0 path and the parent-member path are behaviorally identical to before; `ApplyDialogConnector`'s join/apply mode switch is consistent with the new values.
- **i18n**: all 6 locales change the same single key (`subspaceAdminInvitationsDescription`), valid JSON, key parity preserved, key confirmed referenced via the dynamic `settings.allowedActions.${key}Description` lookup in `SpaceSettingsSettingsView`.

valentinyanakiev:
Re CodeRabbit's nitpick (test coverage for `canAcceptInvitation`): **declined for this PR** — the premise is off: `canAcceptInvitation` is an *unchanged context line* in this diff (the "Changed from parent to current space" comment above it is pre-existing; that change shipped earlier). This PR intentionally scopes its new test to the behavior it changes: privilege-driven `canApplyToCommunity`/`canJoinCommunity` gating. Broader `useApplicationButton` coverage (incl. `canAcceptInvitation`) is a fine follow-up but doesn't belong to this change.

bobbykolev (approved):
Some UX polishments are nice to have but not blocking. 
Commented in the main ticket.

PR #9991, opened 2026-07-08
fix(ux): subspace-link first-visit follow-ups — post-apply landing, cookie-banner dismiss, sign-up return URL

valentinyanakiev:
## Summary

Addresses the three remaining UX issues @bobbykolev reported on alkem-io/server#6189 (the fourth — error toast instead of the parent-first modal — was fixed server-side in alkem-io/server#6232). **Targets `release/66`** so the fixes ship with the combined-subspace-application feature they affect.

### 1. Fresh applicant no longer lands on the parent's Apply CTA
Closing the `/about` route of a private subspace right after applying navigated to the parent space, whose own big "Apply" contradicted the just-submitted application. The close handler now knows the viewer applied this session (`useSpaceApplyFlow` exposes `hasApplied`) and routes a fresh applicant **home**, where the memberships panel confirms the pending application. Browsing-only closes keep today's parent destination; the sidebar About trigger is unaffected.

### 2. Cookie-banner clicks no longer dismiss open dialogs
The consent banner lives outside the Radix dialog layer, so "Accept All" registered as a pointer-down-outside and dismissed the blocking subspace About dialog → redirect to the parent, losing the shared-link context. Fixed generically at the primitive: `DialogContent`/`DialogContentRaw` ignore interact-outside events from any element tagged `data-dialog-dismiss-ignore`; the banner tags its root. Normal backdrop-close preserved, no cookie-specific knowledge in the primitive.

### 3. Sign-up flow preserves the return URL (Signalen-critical)
A first-time visitor on a subspace link who **registered** (vs signed in) landed on home. Three drops, fixed to mirror the working login path:
- login card's sign-up link forwarded no `returnUrl` → now `buildSignUpUrl(returnUrl)`
- the `returnUrl` cookie was session-scoped and died before the verification email was opened → now `maxAge` 1h
- the verification cards' sign-in links called `buildLoginUrl()` bare, which **defaults the param to home and overwrote** any surviving destination → now pass the stored `returnUrl`

**Follow-up (not here):** passing `return_to` into the Kratos registration flow itself (`createBrowserRegistrationFlow`) so Kratos carries the destination through its own redirects — needs allow-list validation; too invasive for a release branch.

### Verification
- `pnpm lint` clean, `pnpm vitest run` 1556 passed, `pnpm build` clean.
- Live browser verification against the local stack to follow in a comment.

Refs alkem-io/server#6189 · `workspace#017-combined-subspace-application`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10000, opened 2026-07-09
Release 66 [patch 1]

bobbykolev:
Patch release on top of Release 66. Six UX/auth fixes, plus an accordion border fix and version bumps to 0.156.2.

The bulk of this is follow-up on #9991 (subspace-link first-visit UX), whose sign-up return-URL fix worked locally but failed in every deployed environment.

---

## 1. Sign-up return URL now survives the identity-subdomain hop

`5db027ecb`

**Root cause.** `CrdLoginPage` normalised the OIDC BFF `returnTo` against `window.location.origin`. In deployed environments `useRedirectToIdentityDomain` serves `/sign_up`, `/registration` and `/verify` from the identity subdomain, and their "Sign in" links are relative — so the check ran on `identity.<domain>` while the stored `returnUrl` was apex-absolute. The origins mismatched, `returnTo` collapsed to `/`, and the BFF returned the user to `/home`. On localhost the identity origin is forced `undefined` in development, so origins always matched and it worked. This fired on **every** deployed sign-up, independent of the email detour.

- Resolve `returnTo` against `platformOrigin` (the apex) via a new `resolveInternalReturnPath` helper in `core/utils/links.ts`.
- New `useSignUpReturnRedirect`: an **armed, single-shot** consumer of the persisted `returnUrl`. Kratos's verification email link is a server-side GET that verifies and 303s straight to `/home` — the SPA's `/verify` never renders and the OIDC request carrying `returnTo` is abandoned. The base-domain cookie survived, but nothing read it. It is armed only when a sign-up carries a destination, disarms on the first authenticated render either way (so it can never hijack a later click on "Home"), and skips auth routes so it can't clear cookies out from under `LoginSuccessPage`.
- **Latent open redirect hardened.** `LoginSuccessPage` did `location.replace()` on the raw `returnUrl` cookie with no origin validation. That cookie is base-domain scoped, so any subdomain could plant an off-platform destination. It is now validated to an in-platform path.
- `buildLoginUrl()` / `buildSignUpUrl()` no longer default the param to `/home`. The login route *stores* whatever `?returnUrl=` it sees, so the default was overwriting a pending destination — and on the identity subdomain it baked `https://identity.<domain>/home` into the cookie. `buildReturnUrlParam` keeps its own default; the `/required` gate depends on it.
- `LoginSuccessPage` now clears the cookie after consuming it (it never did).
- Extracted `isAuthRoutePathname`, now shared with `useOidcSessionRecovery`.

**Not fixed:** opening the verification email on a **different device** still lands on `/home`. No browser cookie can survive that. The fix is passing `return_to` into `createBrowserRegistrationFlow()` so Kratos carries the destination in server-side flow state — it is inert until `selfservice.allowed_return_urls` whitelists the apex in `kratos.yml` (infra repo), so it must land in lockstep. #9991's own commit body flagged this as deferred.

## 2. Reverted the post-apply "land on home" behaviour

`e128cec13`

Reverts the first commit of #9991: closing the subspace About dialog after applying no longer routes a fresh applicant to `/home`. The chain is back to subspace → parent space → home. `useSpaceApplyFlow` keeps `hasApplied` as internal state (it predates #9991 and drives the optimistic disabled Apply button); only its exposure was reverted.

## 3. Apply button removed from the L0 space dashboard

`a244516a5`

The dashboard renders behind `CrdSpaceProtectedRoutes`, which redirects to `/about` when `!canRead`. So on a **private** L0 space a non-member never reaches the dashboard, and a member trips the `isMember` early-return. The button was only ever reachable by a non-member on a **public** space. Removed entirely. The About dialog keeps its Apply button (it calls `useSpaceApplyFlow` directly, not via the connector), and subspaces are untouched.

⚠️ **Behaviour change:** `SpaceAboutApplyButton` also surfaced *Accept invitation*. A user with a pending invitation to a public L0 space no longer sees that CTA on the space dashboard. They can still accept from the home dashboard and the pending-memberships dialog.

## 4. Space-membership setting: copy + placement

`f0511af20`

`allowSubspaceAdminsToInviteMembers` moved to **last** in the Allowed Actions list, and its copy cut from 246 characters to 71:

> **Subspace Membership** — Let Subspaces bring in their own members. They become members here too.

The old label said "**Space** Invitations" for a setting that governs **Sub**space membership, and the description said "this Space", which reads wrong at L1 where the container is itself a Subspace. The new copy is level-neutral. Visibility is unchanged (L0 + L1; `ACTIONS_VISIBLE_AT_L2` still omits it, since an L2 has no child subspaces). All six locales updated with key parity.

## 5. Application-submitted modal explains what happens next

`026bb6319`

Adds a second paragraph with the top bar's bell icon rendered inline via `<Trans>`, so translators control where the icon lands in the sentence:

> Your application will be reviewed. Once it is approved you will get a notification 🔔 and you will be able to enter the community from your Home dashboard.

The paragraph is a sibling of `DialogDescription`, not a child — Radix renders that as a `<p>` and the inline SVG would be invalid markup nested inside it.

## 6. Pending applications on the home dashboard

`31c47ad23`, `817d4a702`

New `ApplicationsBlock` on `DashboardWithoutMemberships`, directly below the invitations, styled as the same rows but with a "Pending" badge instead of Accept/Decline — an application awaits someone else's decision, so there is nothing for the applicant to act on. Renders nothing when empty.

**Canonical visual-fields rule, now enforced and shared.** An L0 space has no avatar concept (its identity image is the cardBanner); an L1/L2 subspace has a real avatar. One is never substituted for the other. This is extracted as `spaceTileImageUrl(level, profile)` in `domain/space/about/utils/` (not `crd/`, which may not import `SpaceLevel`), and it also fixes an existing bug: `mapInvitationsToCards` used `avatar` for every level, so **invitations to an L0 space never showed an image**. Added `object-cover` to `InvitationsBlock` since a wide cardBanner can now land in its square tile.

The backend returns a `Visual` object with an empty `uri` when nothing is uploaded, so absence is a falsy `uri`, not a null object — the mapper coerces it.

---

## GraphQL

`UserPendingMemberships` now selects the AVATAR and CARD visuals on the applications' space profile. Selected **inline rather than in the shared `SpaceAboutMinimalUrl` fragment**: six queries spread that fragment, including `CurrentUserFull`, which runs on every page load. Regenerated output is +27 lines with no unrelated schema drift.

## Testing

`tsc --noEmit`, Biome and ESLint clean. **1602 tests pass** (~50 new). Production build succeeds.

Notable coverage: `LoginCrdRoute.test.tsx` gains the Bug A regression (an apex-absolute `returnUrl` viewed on the identity subdomain) — it **fails on `develop`** and passes here. `useSignUpReturnRedirect` has 11 tests covering the arming, single-shot and ordering-vs-`useOidcSessionRecovery` guarantees. `spaceTileImageUrl` and both mappers are unit-tested per level, including the empty-`uri` case.

## Reviewer checklist — needs manual verification

The identity-subdomain hop and the visuals cannot be exercised on a local stack (all seeded spaces return `AVATAR:empty, CARD:empty`, so the initials tile is the correct local render).

- [x] **Deployed env:** apply to a private subspace as a new user → register → open the verification email **in the same browser** → you land back on the subspace, not `/home`.
- [ ] **No teleport:** log in normally (no sign-up), navigate around, click "Home" within an hour → you stay on `/home`.
- [x] **L0 dashboard:** confirm losing the *Accept invitation* CTA there is acceptable (see §3).
- [ ] **Bare `buildLoginUrl()`:** the error page, the `/verify` unknown-path fallback, and the whiteboard sign-in link now emit `/login` with no param and fall back to the stored cookie.
- [x] **Innovation hub subdomains:** `useSignUpReturnRedirect` treats `/home` as a landing route there too, and bounces an armed user to the apex.
- [x] **Long locales:** the modal sentence is longest in French; the settings description longest in Spanish, in a `max-w-[200px]` caption.
- [x] **Visuals:** how a real 16:9 cardBanner crops into the 40px square tile.

## Known follow-ups (not in this PR)

- Kratos-side `return_to` for the cross-device verification case (needs `kratos.yml` in the infra repo).
- `pendingMembershipsDataMappers.ts` uses `cardBanner` for every level, so `PendingApplicationCard` / `PendingInvitationCard` have the mirror bug for subspaces. Fixing it would *remove* the image from any subspace with a cardBanner but no avatar — a real regression, so it needs a product decision rather than a silent fix.
- Pending applications appear only on `DashboardWithoutMemberships`; a user with ≥1 membership sees none.
- `DashboardWithoutMemberships` reuses `usePendingMemberships`, which also refetches `UserDetails` and the invitation list already fetched by `usePendingInvitationsQuery`. A narrow `communityApplicationsCount` gate would be the right seam.

## Summary by CodeRabbit

* **New Features**
  * Added improved sign-up return handling so users can be sent back to their intended destination after completing authentication.
  * Added safer internal link handling for login and sign-up redirects, preserving valid paths while blocking unsafe destinations.

* **Bug Fixes**
  * Prevented redirect loops and accidental overwrites on authentication and verification pages.
  * Fixed logout and login success flows to clear stored return state more reliably.

* **Documentation**
  * Updated space settings labels and descriptions across supported languages for clearer wording.


bobbykolev:
@coderabbitai review

Issue #9988, opened 2026-07-07
Draft messages to avoid losing them when navigating between spaces while typing a private message


SimoneZaza:
**Describe the bug**
I was writing a long chat message and then moved to a different space, which I expected to be possible because of the "standalone" location of the new chat dialog. However, my entire input field was empty... 

**To Reproduce**

1. Go to any chat
2. Add something in the input field, without sending yet
3. Click to go to the template library or somewhere else
4. See error, input field is empty

**Expected behavior**
The input field to stay populated and the chat to remained opened. I expect this for 2 reasons: 
- Location and UI of the chat, it looks like it's floating and thus not depending on what happens in the 'background
- Current behaviour of other chat platforms, e.g. teams keeps the message in the input field, even when closing teams.

**Screenshots**

**Additional context**

**Areas that will be affected**
_Optional, to be added by developer if relevant_

ccanos:
### Conclussions after investigation
The page reloading is a deeper problem that we may want to look at, but has high risk of having some side effects that we don't want to test/fix now.
Instead of fixing page reloading we are going to do some draft-storing, as the user types the message, the client will be saving in browser's localStorage: conversationId + draftMessageContent.
If the page reloads because of navigation, or the user closing the browser/tab, or the tab hangs or crashes, or the user navigates back to conversation's list, or the user closing the chat popup... which are also causing the message to get lost, the draft will be saved in the browser's memory in all these cases.

### Problems and assumptions
- localStorage is shared across users on the same device. We are talking about private chat messages. We don't want user A to save a draft, then logout, then user B to log in and have access to those drafts.
   We have 3 options here:
   A - localStorage save "userId + conversationId": conversations cannot be mixed accidentally in the UI, but experienced users can access the drafts of other users in the developer tools. Messages survive logouts and tab closes, more or less like MS Teams does.
  B - Clear out drafts on logout: more secure, users can share devices. But it's not explicit and can be confusing to people to lose their drafts on logout. 
  C - sessionStorage instead of localStorage: The drafts are only saved in the current tab session, which also clears out when loging-out, this is the most secure, we don't have to worry about deleting drafts on logout, users can share devices, but different windows and tabs **don't** share drafts between them. It makes the drafts more "ephemeral" which can be more intuitive.
I've assumed **B**

- Orphan drafts: Conversations can be cleared, deleted... Will also delete the drafts in them.
[...]

PR #10089, opened 2026-07-28
fix(chat): keep unsent chat drafts (#9988)

ccanos:
## The problem

An unsent chat message was lost as soon as the composer went away — closing the panel, going Back to the conversation list, or navigating to another page.

Navigation is the reported case and the harshest one. CRD renders in-app links as plain `<a href>` with no router interception, so moving to another space is a **full document navigation**: the SPA reboots and React unmount effects never run. Saving the draft on unmount would have missed exactly the scenario the issue was filed about — so drafts are written **as the user types** instead.

## What changed

- **`src/main/userMessaging/messagingDrafts.ts`** (new) — one `messagingDrafts` localStorage entry shaped `{ [userId]: { [conversationId]: draft } }`. Reads never throw (malformed or disabled storage reads as "no drafts") and drop whitespace-only entries.
- **`src/main/userMessaging/ConversationDraftsContext.tsx`** (new) — drafts in React state above the panel, so they outlive the composer and the conversation list can read them. Persists 500 ms after the last keystroke, flushed on `pagehide` and `visibilitychange → hidden` so the debounce tail survives the navigation. Every write re-reads storage first, so a second tab (or another user of the same browser profile) is never clobbered. `QuotaExceededError` is reported to APM and the save skipped — typing is never interrupted.
- **`CommentInput`** — optional controlled `value` / `onValueChange`. With them the consumer owns the text *and* the clearing; without them the component behaves exactly as before, so the four other consumers are untouched.
- **`UnifiedChatPanelConnector`** — plumbs the draft keyed on the selected conversation, and clears it **only after the send actually succeeded**.
- **`ChatConversationList`** — a conversation with a draft shows `Draft:` plus the text in place of the last-message preview. Key added to all six locales.
- **`LogoutPage`** — drops the whole `messagingDrafts` entry on both logout legs.

## Also fixed along the way

`useConversationView.handleSendMessage` swallows the error and returns `false`, while `CommentInput` cleared its text unconditionally — a failed send silently ate the message. Controlled mode fixes this: the text stays in the composer when the send fails.

The draft is keyed per conversation, which also fixes an unsent draft bleeding into the next thread when the guidance conversation is re-pointed after a clear (the selection changes without remounting the thread view).

## Decisions

- **No TTL and no cap** on stored drafts — the entry is deleted when the draft empties; orphans from deleted conversations may linger, which is fine.
- **Drafts do not change list ordering** or the timestamp; they only replace the preview.
- **Multi-tab is last-writer-wins**, deliberately.
- Keeping the panel *open* across navigation was explicitly out of scope.
- The underlying full-page-reload navigation is a much larger, app-wide change (a delegated click handler or a router-aware link injected into CRD) and was judged too risky to bundle here.

## Testing

`pnpm lint` clean; `pnpm vitest run` → 249 files / 2100 tests passing, 20 of them new:

- storage round-trip, malformed input, quota propagation, logout clear
- provider hydration (own user only), debounce timing, per-conversation isolation, whitespace drop, immediate clear-on-send, other users' drafts preserved, `pagehide` flush
- list rendering: draft replaces the preview, is labelled, and does not reorder

Not yet exercised against a live backend.

https://claude.ai/code/session_01C6yYgDPBLJgoei8bj5aaU2

## Summary by CodeRabbit

* **New Features**
  * Added automatic saving and restoration of unsent chat drafts per conversation.
  * Conversation lists now show draft previews with localized “Draft” labels.
  * Drafts persist across panel closures and page reloads, and are cleared after successful sending or logout.
  * Message composition retains text when sending fails.

* **Tests**
  * Added coverage for draft persistence, synchronization, display, ordering, cleanup, and error handling.


polibb (approved):
looks clean

Issue #10082, opened 2026-07-25
chore: Phase 3 — shell-less client runtime (env-injection out of container startup)


valentinyanakiev:
**3a** of epic [alkem-io/infrastructure-operations#2540](https://github.com/alkem-io/infrastructure-operations/issues/2540) — Distroless Migration Phase 3 (deployment-design tail).

> Previously tracked under alkem-io/client-web#10080 (closed) and briefly under alkem-io/infrastructure-operations#2499. #2540 is the tracking point.

Gated on alkem-io/infrastructure-operations#2452 (port 80 → 8080 + securityContext, release-coupled to client-web v0.160.0) proving out in real traffic.

## Current blocker for shell-less
The container's CMD runs `/bin/sh -c "env.sh && nginx"` — runtime SPA env-injection requires a shell inside the image.

## Ask
- Move env-injection out of container startup. Candidate approaches (pick one, record the decision):
  1. initContainer writes `env-config.js` to a shared emptyDir volume (keeps runtime image inert)
  2. build-time templating per environment (loses run-anywhere image, simplest runtime)
  3. nginx `sub_filter`/njs injection (no extra container, more nginx config)
- Then adopt a shell-less non-root nginx-class base (e.g. Chainguard nginx) or a minimal static file server, digest-pinned.
- Update the Phase 1 smoke script: additionally assert no `/bin/sh` in the runtime image.

## Acceptance
- [ ] No shell required at container start; env-injection verified in a live environment
- [ ] Runtime image has no shell/package manager; scan clean; size recorded
- [ ] Smoke script updated and green

valentinyanakiev:
**Status: not started, and deliberately so.** Phases 1–2 are implemented and in review (alkem-io/client-web#10095, alkem-io/dev-orchestration#71); this phase was explicitly held back.

Recorded as Clarification C-1 in the workspace spec (`033-client-runtime-hardening`, alkem-io/agents-hq#78): Phase 3 is a stretch goal **gated on Phases 1–2 proving out in real traffic**, and its env-injection replacement needs a considered architectural decision rather than a rushed one. Forcing the choice now would be premature.

### What Phase 1 established that changes this phase's shape

The runtime shell is still required at container start: `CMD ["/bin/sh", "-c", "/usr/share/nginx/html/env.sh && nginx -g 'daemon off;'"]`. Phase 1 confirmed **exactly what `env.sh` does at runtime**, which the three candidate approaches must all reproduce:

[...]

Issue #10080, opened 2026-07-25
Client Runtime Image Hardening (non-root nginx → shell-less)


valentinyanakiev:
## Description

Harden the client-web production container — currently the **least hardened image in the Alkemio fleet**: a floating (unpinned) `nginx:alpine` base, running as **root**, with a full shell and package manager, and a startup shell script (`env.sh`) the container depends on.

Sibling epic: alkem-io/infrastructure-operations#2059 (Secure and Minimal Artifacts / Distroless Migration) covered every backend service; client-web was excluded there because an nginx-served SPA needs a different pattern — this epic is that pattern, phased so each step ships independently.

### The job to be done
- As a security engineer, I want the public-facing web entrypoint to run non-root on a pinned base (ISO 27001 A.8.31; the same posture every backend service now has) so a compromised container has minimal pivot surface.
- As an operator, I want deterministic builds (digest-pinned bases) and standard `runAsNonRoot` enforcement in the manifests.

## Phases

**Phase 1 — pinned, unprivileged nginx (image)**
Swap to `nginxinc/nginx-unprivileged` (UID 101, listens 8080), digest-pin builder + runtime, keep `env.sh` (image retains a shell — non-root is the goal of this phase, not shell-less). Persisted smoke script + scan evidence per the 026 convention.

**Phase 2 — manifest rollout (coupled)**
Service `targetPort` 80→8080, `runAsNonRoot`/no-escalation/drop-ALL securityContext, probes — in dev-orchestration (dev/test/sandbox) and infrastructure-operations (acc/prod, release-coupled). **Apply only in the release that ships the Phase 1 image** (same rollout-ordering rule as workspace 026).

**Phase 3 — shell-less runtime (stretch)** — **moved out of this epic 2026-08-05**
Phase 3 (#10082) now lives under the follow-up epic alkem-io/infrastructure-operations#2499 (Distroless Migration — Wave 2), alongside the rest of the fleet's remaining distroless work. **This epic closes on Phases 1–2**, which are delivered: hardened image + manifests live on dev/test/sandbox as of 2026-08-04 (workspace#033; acc/prod rollout rides the release-coupled infra-ops change tracked in Wave 2).

## Reference
Pattern, evidence conventions, and rollout-ordering rationale: workspace `026-distroless-runtime-images` (alkem-io/agents-hq#75).

valentinyanakiev:
## Epic status 2026-07-25 — Phases 1–2 delivered, Phase 3 outstanding

[...]

PR #10095, opened 2026-07-29
chore: Phase 1 — pinned, unprivileged (non-root) nginx runtime image

### LINKED ISSUE #10081: chore: Phase 1 — pinned, unprivileged (non-root) nginx runtime image
Phase 1 of epic #10080 (Client Runtime Image Hardening).

## Current state
Final stage `FROM nginx:alpine` — floating tag, master process runs as **root**, binds port 80; `CMD` chains `/bin/sh -c env.sh && nginx`.

## Ask
- Runtime base → `nginxinc/nginx-unprivileged:<pinned-version>-alpine`, **digest-pinned** (builder `node:24.14.0-alpine` digest-pinned too).
- Container listens on **8080** (unprivileged), runs as UID 101; `env.sh` keeps working (this image retains a shell — shell removal is Phase 3).
- Adjust nginx.conf as needed for the unprivileged layout (pid/temp paths differ from stock nginx).
- Persisted smoke script (026 convention): asserts non-root UID, pinned digests, listens 8080, SPA index served, `env-config` injection still works.
- Trivy scan + size recorded; zero fixable HIGH/CRITICAL.

## Acceptance
- [ ] Image runs as UID 101, serves the SPA on 8080 locally (docker run, no root)
- [ ] Both bases digest-pinned
- [ ] Smoke script committed and green; scan evidence recorded
- [ ] NOTE for release: manifests must switch targetPort/securityContext in the same release — Phase 2 (do not deploy this image against old manifests)

valentinyanakiev:
Closes #10081 — Phase 1 of epic #10080 (Client Runtime Image Hardening).
Workspace spec: `agents-hq/specs/033-client-runtime-hardening` (alkem-io/agents-hq PR pending).

## Why

client-web was the least hardened image in the fleet: a **floating** `nginx:alpine` tag running as **root**. Every backend service was hardened in workspace 026; this closes the client gap.

## What

- **Both stages digest-pinned** (were `node:24.14.0-alpine` and `nginx:alpine`, both floating).
- Runtime → **`nginxinc/nginx-unprivileged:1.30-alpine`**, UID **101**, `EXPOSE 8080`.
- `nginx.conf`: pid + temp paths to `/tmp`, `listen 8080` — a non-root process can neither write `/var/run/nginx.pid` nor bind a privileged port.
- Web root chowned to 101: `env.sh` **writes** `env-config.js` and `robots.txt` there at startup, so the unprivileged user must own it. This is the one substantive behavioural difference from the root-era image.
- `env.sh` / `.env.base` injection otherwise **unchanged** — removing the runtime shell is Phase 3 (#10082).
- New persisted regression `.docker/client-image-smoke.sh` (9 assertions: digest pinning, non-root at rest *and* at runtime, 8080-only, SPA + deep-link serving, env-injection honouring container env).
- README `docker run` port corrected to 8080.

## Evidence

| | HIGH+CRITICAL | fixable |
|---|---|---|
| `nginx:alpine` (before) | 26 | **26** |
| `nginx-unprivileged:1.29-alpine` (first candidate) | 12 | 12 |
| `nginx-unprivileged:1.30-alpine` (**shipped**) | **0** | **0** |

Rationale for the base choice in `.docker/verification/US1-AS4-cve-assessment.md`; SBOM + trivy JSON in the workspace evidence bundle.

Gates: `pnpm lint` clean, **246 test files / 2083 tests pass**, image builds, smoke script green.
Live-verified on `k8s-hetzner-sandbox` in an ephemeral namespace: pod Ready non-root (kubelet-enforced `runAsNonRoot`), SPA + deep links 200 **through the Service**, runtime env-injection working unprivileged, `id -u` → 101 in-cluster.

## ⚠️ Rollout ordering — do not deploy alone

Requires the matching manifest change (**alkem-io/dev-orchestration#71**: `targetPort` 8080 + securityContext) **in the same release**. This image against the old port-80 manifests serves nothing; the old root image against the new manifests fails to start. Acceptance/production manifests (infrastructure-operations) are a separate release-gated follow-up — workspace#033 R-1, tracked on alkem-io/infrastructure-operations#2452.

## Summary by CodeRabbit

* **Security**
  * Hardened pods and containers to run as non-root (UID/GID 101), drop all capabilities, and apply stricter security settings.
  * Digest-pinned base images and switched to an unprivileged runtime for improved supply-chain and runtime safety.
* **Operations**
  * Updated liveness/readiness probes to use HTTP GET `/` on port **8080**.
* **Documentation**
  * Updated Docker run instructions to port **8080** and clarified that Kubernetes Services continue exposing port **80** while forwarding to the container’s **8080**.


valentinyanakiev:
## Review remediation (2026-07-25)

An adversarial review round returned a **security hold** on this branch, and it caught a real defect. Four additional commits since the PR opened:

### The miss that mattered: this repo's own manifests are the live apply path

The workspace spec named `dev-orchestration` as the only Phase 2 consumer. That was wrong — **client-web ships its own `manifests/{25,26,30}-*.yaml`, applied by `build-deploy-k8s-{dev,test,sandbox}-hetzner.yml` via `azure/k8s-deploy` on every `develop` push**, and those are demonstrably what author the running Deployments (verified live: they carry `workflow=actions.github.com-k8s-deploy` labels and lack the kustomize base's affinity/resources).

Merging this PR with those manifests stale would have routed Service:80 → pod:80 (nothing listening) and **502'd dev** — and would have silently re-applied `containerPort: 80` with no securityContext over any correct kustomize apply. All three now carry the full Phase 2 fields, verified live on sandbox with these exact manifests.

### Security fixes beyond the original scope

- **`automountServiceAccountToken: false`** — the default ServiceAccount is bound by `ClusterRoleBinding/pod-reader` to `pods/exec` and `pods/log` **cluster-wide** (verified against k8s-hetzner-dev). Mounting that token into the internet-facing SPA pod was a free privilege-escalation step. The SPA never calls the Kubernetes API.
- **Dropped `envFrom: secretRef: alkemio-secrets`** — the platform-wide Secret (AI provider keys, Matrix tokens, signing keys) was being injected into a public static-file server that uses none of it.

### Rollout safety: named `targetPort`

`targetPort: web` rather than the numeric `8080`. Kubernetes resolves a named targetPort **per-pod**, so during the rollout old pods (`web`=80) and new pods (`web`=8080) are both served through the one Service. Proven on sandbox: one Service, two EndpointSlices (`:80` and `:8080` simultaneously), **20/20 requests 200** during the mixed state. The numeric form would have blackholed every old pod on apply.

### Regression protection

`.docker/client-image-smoke.sh` is now wired into `ci-build.yml` (on `pull_request`), so the digest-pin / non-root / 8080 invariants can't regress silently. It also measures cold start mechanically (`STARTUP_MS=`, currently 279 ms) and fails above the 3s readiness delay — the probe timings were previously justified only by prose.

PR #10108, opened 2026-08-04
Show join/apply button on top-level space pages

### LINKED ISSUE #10105: BUG: no way to apply/join from the welcome-space page as a non-member
**Describe the bug**
As a user who is not a member of any spaces, when I navigate to the welcome-space (a public space that requires an application to join), I don't see any way to apply/join from that page. There is no visible join/apply entry point.

**To Reproduce**

1. Log in as a user who is not a member of any spaces.
2. Navigate to the welcome-space (public, application-required space).
3. Look for a way to apply/join the space.
4. No join/apply control is visible.

**Expected behavior**
A user viewing a public, application-required space they're not a member of should see an entry point to apply to join (e.g. an Application button), as has historically been the case elsewhere in the app.

**Additional context**
The classical queries that have always backed this button — `ApplicationButton` and the `RoleSet` query — don't appear to be present/wired up on this page. This may indicate the join/apply flow has been broken for this entry point in client-web for some time, though that should be verified rather than assumed — it's surprising this would have gone unnoticed.

Needs investigation to confirm whether the `ApplicationButton`/`RoleSet` queries are genuinely missing from this page or just not rendering/wired correctly, then fix the missing join/apply entry point on the welcome-space page.

**Areas that will be affected**
_Optional, to be added by developer if relevant_

polibb:
## Summary
- `SpaceApplyButtonConnector` (the join/apply button, backed by the `ApplicationButton`/`RoleSet` data) was only mounted in `CrdSubspacePageLayout` — the top-level space layout (`CrdSpacePageLayout`) never rendered it, so non-members visiting a public, application-required root space (e.g. welcome-space) had no way to apply or join.
- Wire `SpaceApplyButtonConnector` into `CrdSpacePageLayout`, mirroring the subspace layout: hidden while loading/already a member, hidden on the settings pages, with the same `mb-6` spacing the subspace layout gets from its `space-y-6` wrapper (this layout doesn't have one).

image

Fixes #10105

## Test plan
- [x] Added `CrdSpacePageLayout.test.tsx`: TDD'd against the bug — written first, watched it fail (button not rendered), then made it pass with the fix. Covers rendering with correct props and hiding on the settings route.
- [x] `pnpm vitest run` — full suite green (pre-existing bg_BG locale flake in `budgetMeter.test.tsx` unrelated to this change)
- [x] `pnpm lint` — clean

## Summary by CodeRabbit

- **New Features**
  - Added an Apply/Join button to public root-space pages.
  - The button displays relevant space information and appears before the page content.
  - The button is hidden on space settings pages.

- **Tests**
  - Added coverage verifying the button’s visibility and displayed details across supported views.

### DIFF
--- a/src/main/crdPages/space/layout/CrdSpacePageLayout.test.tsx
+++ b/src/main/crdPages/space/layout/CrdSpacePageLayout.test.tsx
@@ -1,0 +1,155 @@
+import { render, screen } from '@testing-library/react';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { SpaceLevel } from '@/core/apollo/generated/graphql-schema';
+import CrdSpacePageLayout from './CrdSpacePageLayout';
+
+// ---- Mocks ----
+// CrdSpacePageLayout is a large orchestrator with many data/UI hooks. Everything
+// unrelated to the join/apply wiring under test is stubbed to a minimal value so
+// the component can render without a live Apollo/router tree, following the
+// module-mock pattern used by CrdSubspaceProtectedRoutes.test.tsx.
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (key: string) => key }),
+}));
+
+let mockPathname = '/welcome-space';
+vi.mock('react-router-dom', () => ({
+  Outlet: () => <div data-testid="outlet" />,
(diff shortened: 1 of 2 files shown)


polibb:
removed on purpose in regards to Signalen's usage of the platform

PR #10175, opened 2026-08-21
feat(051): sign-up signpost — warn a provider-arriving visitor they may already have an account

hero101:
## What

Adds a **sign-up signpost**: when a visitor arrives at the sign-up screen mid-way through signing in with an identity provider (an OIDC-continued registration flow), the page now warns them they may already have an Alkemio account and offers a working route to sign in instead — before they silently create a duplicate account.

Direct email/password registration is untouched: the signpost renders **only** when the loaded registration flow evidences a provider continuation, and the pre-existing "account already exists" hard-redirect keeps precedence.

## Why

A person who already has an Alkemio account and signs in with a provider identity that isn't connected to it cannot be matched to their existing account (provider email need not match). Today they silently acquire a second account — a live defect. This signpost is the discovery path, shipped ahead of the wider Connected Accounts work.

Part of alkem-io/alkemio#2079 · epic alkem-io/alkemio#1947 · workspace#051-cleverbase-account-linking

## How

- `SignUpCrdRoute`: provider-arrival discriminator (`flow.active === 'oidc'` + contracted fallback evidence), flow-scoped sessionStorage origin marker so a deliberate provider registration doesn't false-fire the warning; the `4000007` redirect branch keeps precedence.
- `SignUpCard`: plain in-flow signpost slot (warning + sign-in link, WCAG 2.1 AA; link built via the existing internal-path resolver — no new redirect sink).
- i18n: `signUp.signpost.*` in all six languages (parity-tested).

## Evidence digest

- Unit/component: `SignUpCrdRoute.test.tsx` (+ new cases: signpost for OIDC flow, absent for plain flow, redirect precedence, stale-marker non-suppression), `SignUpCard` markup covered directly; `auth.parity.test.ts` green.
- Live: direct-registration negative leg verified on an isolated stack (no signpost on email/password sign-up); positive provider-arrival leg requires a completed external OIDC hop — covered by unit tests now and by the acceptance quickstart walk (V2) before story sign-off.
- Review: adversarial multi-round panel (correctness / spec-compliance / quality / security); all confirmed findings on this branch fixed. **Security verdict: pass** — no enumeration oracle, no open-redirect sink, flow-scoped marker cannot suppress a different flow's warning.
- Gates: `pnpm lint`, `pnpm vitest run` (full suite), `pnpm build` — green; commits signed.

## Tracked debt (low, non-blocking)

- A spent Kratos flow id remains in tab-scoped sessionStorage on the successful provider-registration path.
- The shipped three-input discriminator deviates from the initially pinned two-row contract table; recorded for a doc sync.

## Summary by CodeRabbit

- **New Features**
  - Added a sign-in prompt for people arriving at registration through an existing account provider.
  - The prompt includes a clear link to continue to sign-in.
  - Intentional provider-based sign-ups no longer show the unnecessary sign-in prompt.
- **Localization**
  - Added and updated sign-in warning and privacy-policy text in English, Bulgarian, German, Spanish, French, and Dutch.
- **Bug Fixes**
  - Improved handling of provider sign-up and sign-in flows, including stale or incomplete flow information.


hero101:
Closing as consolidated into #10176.

**Why:** every commit on `feat/051-cleverbase-account-linking-signpost` is already contained in `feat/051-cleverbase-account-linking` (head `feeea3e6` is an ancestor of that branch), so this PR's four commits are part of #10176 and would otherwise be reviewed twice.

**What changed since:** the signpost has been substantially reworked following a live acceptance walk, so the version under review here is superseded rather than merely duplicated:

- The warning now leads with what continuing actually does when the provider email matches an existing account — the provider is connected to that account and both sign-in methods stay usable — instead of a generic "you may already have an account".
- The sign-in route is relabelled "Sign in with a different account". Following the old route with the *same* provider looped back to the sign-up screen, because an unlinked identity is bounced to registration; the route genuinely serves only the case where the existing account uses a *different* email, which is also the only case that can silently create a duplicate.
- Every provider continuation (not just a signposted one) now explains the confirm-your-details form: a heading naming the action and provider, a sentence on where the prefilled values came from, and a labelled primary action instead of a bare provider glyph.
- Copy stays conditional throughout — this screen renders before any server verdict on the email and must not confirm whether an address is registered.

Both surfaces, their gating, and the final copy are reviewable in #10176. No work is lost by closing this.

PR #10238, opened 2026-08-28
fix: recover collaboration without blocking established editors

antst:
Free-text bug, no board story.

## Root cause

The lifecycle patch treated transient transport loss as if an established document had become unavailable: it reset collaboration/scene readiness, consumers combined raw connection flags into edit locks and loading/modal state, and memo unmounted its editor. Separately, the provider killed any initial sync after a fixed 15 seconds, so a progressing template source could never finish before the outer loader failed.

A whole-flow review also found that dialog close was still outside the provider durability contract, durability could expedite a terminal connection as if it were an explicit retry, memo admission replacement briefly removed its provider, and transport lifecycle was conflated with read-only access.

## Fix

- Keep one provider-owned connection/retry scheduler and replace fixed initial-sync/echo timers with inbound-idle liveness: valid inbound progress resets the deadline, then type-5 heartbeat gets a grace period before the normal transient-close funnel.
- Keep the established editor and same Y.Doc/Scene mounted through transient recovery. Memo admission replacement atomically swaps providers over that same Y.Doc.
- Derive transport lifecycle and server access as orthogonal facts shared by memo and whiteboard. Recovery remains non-blocking; read-only access alone locks editing.
- Track local updates monotonically. Provider-owned `persistPendingChanges` drains every pending update, including a newer edit that lands while an older durability acknowledgement is in flight.
- Freeze close input, then await import/assets/provider persistence before teardown. Draft whiteboards force a barrier; ordinary clean documents avoid a needless wire request. Failed persistence keeps the editor open. Delete intentionally bypasses normal persistence.
- Separate internal durability acceleration from explicit Retry. A terminal or intentionally disconnected provider cannot be revived by a barrier waiter; explicit Retry still uses the one scheduler.
- Seal terminal whiteboard ownership synchronously and keep dirty terminal scenes visible/exportable. Later socket status cannot clear terminal state.
- Make template import additive exactly once and report success only after the target room confirms persistence.
- Preserve #10236's intended prompt offline detection, inactivity Resume action, caught rename errors, hard asset/Yjs close gates, and best-effort preview/profile metadata.

## Regression coverage

- Provider: intentional destroy ordering; remote 1000 reconnect; terminal no-retry/no-durability-redial; explicit Retry; offline/in-flight dial guards; explicit v0.0.2 type-5 echo; sleep/wake grace; progressing sync beyond 15 seconds; silent-stall close; old reply correlation; SYNC-before-dirty-listener ordering; stale acknowledgement requires a second barrier.
- Lifecycle/access cross-products: initial/live/recovering/terminal/generation replacement remain independent from read-write/read-only access.
- Whiteboard: transient recovery stays mounted and non-blocking; terminal is sealed/disconnected; dirty terminal scene stays visible; normal and draft close persistence; duplicate close ignored; assets are included; persistence failure stays open; preview failure remains best effort.
- Memo: transient recovery and same-Y.Doc provider replacement mount once; close freezes and persists; duplicate close ignored; newer edit needs a second acknowledgement; failure stays mounted/editable; delete bypasses persistence.
- Template source progress, merge-once/target-persisted semantics, and two explicit additive imports remain covered.

## Gates

- `pnpm run lint` — green (`typecheck:native`, Biome 2,005 files, ESLint; no errors)
- Pinned Node 24.14.0 full Vitest — green (362 files; 3,110 passed, 2 skipped)
- Focused lifecycle/whiteboard/memo Vitest — green (25 files; 232 passed), plus final close/delete pins (19 passed)
- `pnpm run test:e2e:typecheck` — green under Node 24.14.0
- `pnpm run build` — green under Node 24.14.0 (7,818 modules)
- Normal signed pre-commit hook — green (lint-staged, TypeScript, ESLint, full Vitest)

The shell's Node 26 run failed only unrelated jsdom/localStorage tests because that runtime exposed no localStorage. Re-running an affected suite and the complete suite under the repository-pinned Node 24.14.0 was green; no product workaround was added for the runner mismatch.

## Hard pre-merge evidence blockers

This PR is intentionally draft until both are captured on acceptance without mutation:

1. A WebSocket trace for the reported draft/reconnect loop showing the explicit type-5 echo and actual close code/reason, so the original acceptance failure is attributed rather than masked.
2. The failing source template's first-sync byte size and duration, so the final caller deadline is evidence-based. The current 120-second cap is provisional; provider liveness, not that cap, governs progressing sync.

Authenticated local browser fault injection was not claimed: the isolated Vite origin did not own the authenticated local gateway/session, and taking over the user-owned origin would have disturbed another worktree.

No SpecKit artifact was created. Added/updated comments contain no spec or issue references.


antst:
Superseded by #10239. The replacement cleanly reverts #10236 and rebuilds the collaboration lifecycle around one per-document provider state, one reconnect owner, one post-ready heartbeat, and one single-flight durability operation. This branch is preserved as investigation evidence but will not be extended.

PR #10306, opened 2026-09-14
fix: give signed-copy details and actions room to breathe

antst:
## Source

Free-text bug, no board story. User reported the signed-success form was jammed: “Everything is allright except one form” and “I think this dialog can be a bit iproved to be less jammed”, with a screenshot showing truncated signer metadata beside three actions.

## Root cause and fix

`SignedCopy` switched to a horizontal flex row at the viewport `sm` breakpoint even though its success dialog remains narrow. The three non-shrinking action buttons took the space needed by the signer's name and timestamp. At a 1280×720 viewport, the actual 576px dialog left the signer just 35.6px wide.

Remove that competing horizontal layout: keep metadata on its own row above a wrapping toolbar, allow signer names to wrap, and increase the action gap to 8px. The existing primary Open signed PDF and quieter Download / Verify actions are unchanged. The same card in signed-copy history also stays readable.

No widening the dialog, smaller text, new responsive machinery, dependencies, translations, or modal/signing/auth/query/storage behavior. No comments changed; no spec artifact.

## Verification

- RED on develop `ac5e1bcb2d012f2971e718f54681b5354924782c`: signing/history layout cases fail before the fix (2 failed / 38 passed). These are layout-contract tests, not claims of rendered geometry.
- GREEN focused: 40/40 tests.
- Full Node 24 suite: 403 files, 3,471 passed / 2 skipped.
- Production build, full lint (137 existing warnings), typecheck, Biome, React Compiler lint, and diff-check pass. Normal signed-commit hook, no bypass.
- Real Chromium renders of the actual component and CRD/global CSS, with Source Sans Pro loaded and entry animations complete: 1280×720 narrow success dialog, 375×667 mobile, Dutch long-name/verified result, Bulgarian mobile long-name/verified result, and history. All five have untruncated names, a toolbar below metadata, no overlapping buttons, and no horizontal card overflow. Desktop metadata grows from 95.6px to 492px wide, with a 16px vertical gap before actions.
- Before/after screenshots and measured rectangles retained locally under `test-results/signed-copy-spacing/`; user-facing preview copies under Agents HQ `out/signed-copy-spacing/`. This is standalone presentational browser proof, not an authenticated signing-flow or deployment claim.

Human merge; no deployment or board transition performed.

## Summary by CodeRabbit

* **Style**
  * Improved the memo signing dialog layout on narrow screens.
  * Signer names now wrap instead of being truncated.
  * Action controls have more spacing and wrap independently.
  * Signed-copy metadata is displayed vertically for better readability.


antst:
@coderabbit review

bobbykolev (approved):

PR #10329, opened 2026-09-18
feat(070): off-by-default "Notify space members" switch when adding a response or task

bobbykolev:
## What & why

Adds an off-by-default **"Notify space members"** switch to the response/task creation dialog, so a board or a Post can be worked on without filling everyone's inbox. Client half of `alkem-io/notifications#593`.

## Scope

Exactly one contribution-creation component changes: `CrdPostContributionDialog.tsx` — which already serves **both** the "response" and the "task" paths (behind `PostContributionAddConnector` / `TaskBoardConnector`), so this is one dialog, not two surfaces. Plus its own `.graphql` document, the regenerated codegen output, and its test.

- **CRD only** — `src/crd/*` primitives (`Switch` / `Label`); the legacy MUI tree is untouched.
- **Zero new i18n keys.** Reuses the existing `crd-space` `forms.notifyMembers`, already present in all six locales (en/nl/de/fr/es/bg) and already used by the shipped callout-creation switch. The architect council's dissent D3 argued for a context-specific key and was overruled in favour of consistency; minting one later is a one-line follow-up.
- **Create mode only.** Edit mode sends nothing.
- **Never remembered** — the switch is per-dialog form state and returns to OFF on every open, including across callouts.
- No Apollo cache-shape change, no data changes.

The other four response connectors (link / memo / whiteboard / document) show **no** switch and are untouched — they keep notifying as today by omitting the field.

## Why off-by-default lives here and not in the schema

The server's wire default stays `true`; "off by default" (AC2) is client form state. Putting `false` in the contract would have silently inverted behaviour for every non-client API producer. The client sends an explicit `false` only when the user leaves the switch off.

## Evidence

Built with `/forge`. Client unit tests **10/10** (`CrdPostContributionDialog.test.tsx`), lint and build green.

Verified live in a browser against a running stack — US1 all six scenarios pass, with screenshots:
- the switch renders and is **OFF** on every open, for both the response dialog and the Tasks-board task dialog;
- it is never remembered, even after a prior submission with it ON;
- submitting with it OFF emits **zero** notification events, while the response still appears in the callout and in Recent Activity identically to a notified one;
- a recipient who explicitly opted **in** still receives nothing when the sender leaves it off.

US2 (switch ON notifies as today; draft callouts still silent; opt-out never overridden) also passes.

**Review**: security verdict for this repo **conditional** — no critical or high findings open. No client-side findings required a code change.

## Tracked residual

Worth knowing as a reviewer, though it is a server-side and product decision, not a client one: with the switch off, **neither** the community **nor** the space-admin notification is sent (AC3 honoured literally, per intake ruling D-OP2 over a recorded architect-council dissent). The activity feed remains the complete record. Residual-risk acceptance is a human release-gate decision.

Rollback ordering matters: if this feature is reverted, **revert the client first** — a deployed client sending `sendNotification` to a reverted server would fail GraphQL validation.

## Links

- Story: alkem-io/notifications#593
- Workspace feature: `workspace#070-contribution-notify-switch`
- Spec / plan / evidence ledger: `agents-hq/specs/070-contribution-notify-switch/`
- Sibling PRs: alkem-io/server#6524, and test-suites (same branch name)

## Summary by CodeRabbit

- **New Features**
  - Added a “Notify members” option when creating posts.
  - The option is off by default and can be enabled before publishing.
  - Enabling it notifies members when the post is created.
  - The setting resets when the post dialog is reopened.
  - The option is available for responses and task-board tasks, but not when editing posts.


bobbykolev:
Both review comments addressed in `ea9f5d80a`.

### 1. Codegen against the live endpoint — done, and the committed output is confirmed correct

Regenerated with `CODEGEN_SCHEMA` pointed at the running feature server. **Result: the committed generated output is unchanged** — `pnpm codegen` against the live schema produces byte-identical files.

Validation summary:

| check | result |
|---|---|
| live introspection of `CreateContributionOnCalloutInput.sendNotification` | present, `defaultValue: "true"` |
| `CreateCalloutContributionData` (template/replay type) | **no** `sendNotification` field — the flag did not leak onto the shared base class |
| generated `CreatePostOnCallout` operation | `($calloutId: UUID!, $post: CreatePostInput!, $taskColumn: String, $sendNotification: Boolean)`, passed inside `contributionData` |
| diff vs committed output | none |
| server schema-diff CI (on the sibling server PR) | **0 breaking**, 34 additive, no blocking changes |

One wrinkle worth recording, since it would otherwise look like drift to the next person: generating from the live **introspection** endpoint drops the scalar doc-comments in the `Scalars` block (`DateTime`, `Emoji`, `NameID`, …) that generating from the printed SDL preserves. `origin/develop`'s committed file carries those comments, so a naive live regeneration produces a spurious 9-line deletion unrelated to this PR. That deletion was reverted; the feature's own types are identical either way.

### 2. Switch should mark the form dirty — fixed

Valid, and on this file's own precedent: the `targetCalloutId` `Select` is likewise a non-content choice and explicitly calls `setIsDirty(true)`, while the notify switch was the one control that didn't. So toggling it and then cancelling (or clicking outside) closed the dialog silently instead of routing through the discard confirmation.

`onCheckedChange` now sets `setIsDirty(true)` alongside `setNotifyMembers`, matching that idiom.

Gates: lint clean, `CrdPostContributionDialog.test.tsx` **10/10**.

ccanos (approved):

